PR #20, opened 2019-04-19
Add rain, rain delay and operation sensors

davidjb:
This adds several additional binary sensors for the rain sesnor, rain
delay and whether operation is enabled in general, as well as a sensor
for the Rain Delay Stop Time.

Also features a minor bit of refactoring for querying controller
variables to reduce code duplication when hitting the `/jc` endpoint as
per the API (see
https://openthings.freshdesk.com/support/solutions/articles/5000716363-os-api-documents)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,9 @@ Last tested on OS API `2.1.8` and Home Assistant `0.88.1`
 ### Features
 
 - Binary sensors for each station to show on/off status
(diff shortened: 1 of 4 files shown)


vinteo:
Thanks so much for this, I will do a quick test and if its all good, will merge

vinteo:
Looks good, thanks again for this. Merging...

davidjb:
@vinteo No worries, thanks for making this addon 😃

Issue #24, opened 2019-06-25
Encoding issue


tellerbop:
I have swedish characters in my station names, programs etc.
May I suggest that you add the following to each  request?

`response.encoding = response.apparent_encoding`

this solves the encoding issue for the response

vinteo:
Hi,

Feel free to open a PR as I wouldn't know how to test it fully for your
scenario.

Thank you

tellerbop:
Done

PR #25, opened 2019-06-25
Adding response encoding

tellerbop:
Adding response encoding to resolve response treated as ISO-8859-1 (default) when it should be UTF 8 as detected by the response

response.encoding = response.apparent_encoding detects character encoding. Opensprinkler API doesn't correctly set the charset header Content-Type: text/html; charset=utf-8


vinteo:
Thanks! I will do a quick test tonight to make sure it still works for me before merging.

vinteo:
Fixes #24

vinteo:
Thanks for your PR

PR #37, opened 2019-11-20
Add support for OS 2.19

hwstar:
Added support for Open Sprinkler version 2.19 which added an additional rain sensor, and changed the REST commands to access both sensors.


vinteo:
Thank you very much for this!

PR #52, opened 2020-05-25
OpenSprinkler Integration with HASS UI Configuration

### LINKED ISSUE #6: Multiple devices support
Allow for multiple OpenSprinkler devices

### LINKED ISSUE #27: Add config for scheduled update time
Currently its hardcoded at 5 seconds. Could probably bump it up to 10 secs, but ideally it should be a config option

### LINKED ISSUE #35: Include "remaining water time" in sensors
First off - great work!
One suggestion for enhancement is to add the remaining time to a running station. This way you can approximate even more all the info you get on the native OpenSprinkler UI.

As a quick poc, I looked at the code and you already have that info on the return of the jc call.

      if(self._p_status[0]==99):
        self._state = "Running manual (" + str(self._p_status[1]) + "s remaining)"

This works :)

Adding it to the status text might not be ideal, but maybe a property?
Also, without some custom UI, it would only ever update on the scan interval, and not every second like OS does, but that could be a further enhancement...

### LINKED ISSUE #38: Delayed sensor setup on restart
Hi,

Great work.

As expected, if OpenSprinker is down when Hassio starts up the entities are not set up and are not shown as available. However, Open Sprinkler always takes longer to restart than Hassio, so if there is a power failure I always need to reboot Hassio a while after to get the entities working.

Is there any way to delay the set-up of the entities so that they can be created after Open Sprinkler is running? I'm guessing the answer is no, but thought I'd ask!

Thanks,
Matthew

### LINKED ISSUE #50: Support https
I run opensprinkler behind a reverse proxy. Any chance you can add an ssl option or similar to support encrypted connection?

Thanks!

### LINKED ISSUE #63: Change domain from hass_opensprinkler to opensprinkler
How does it affect HACS?

vinteo:
This is the feature branch for the next major version of the OpenSprinkler Integration

Closes #6 
Closes #27 
Closes #35 
Closes #38
Closes #50 
Closes #63

# TODO

- [x] make services available to binary_sensors entities
- [x] state attributes - https://github.com/vinteo/hass-opensprinkler/issues/56, https://github.com/vinteo/hass-opensprinkler/issues/57, https://github.com/vinteo/hass-opensprinkler/issues/58
- [x] additional sensors (or maybe state attributes) for:
  - [x] waiting station count
  - [x] active station count
  - [x] current running program
  - [x] idle station count
  - [x] total stations
- [x] pull in latest pyopensprinkler
- [x] ~~config option for setting custom `SCAN_INTERVAL` #27 (Not sure if this is needed)~~
- [x] availability
- [ ] support updating the device options (url, password, etc)
- [x] use mac in unique_id 
- [x] simplify config to just be url vs host, port, ssl, etc (this would facilitate running in a proxy with a subdir as well) #71 
- [x] audit icons #72 
- [ ] audit available services


vinteo:
@travisghansen I have copied over what I have in the HASS PR to this branch, we can work off this now

travisghansen:
Should we make `mac` be a required user input to facilitate the unique requirement?

vinteo:
I am still conflicted about that, we would probably need it to get the official integration in but its not very user friendly. Maybe we should try to see if we can get OpenSprinkler to add the mac address to their API

travisghansen:
Agreed but then we have to wait for that to cycle through. We can make it automated/optional at some point but require it for the time being?

vinteo:
Added it in the 0.5.0 PR. 

Can I leave it to you to create an issue in the OpenSprinkler repo?

travisghansen:
Yeah I’ll bring it up over there. They are also working on mqtt support I saw which may tie into this at some point.

travisghansen:
https://github.com/OpenSprinkler/OpenSprinkler-Firmware/issues/124

travisghansen:
So far really good! I think we need to update the logic in the lib to better handle master status.

```
    @property
    def status(self):
        """Retrieve status"""
        is_running = self.is_running
        pid = self.running_program_id

        if is_running:
            if pid == 99:
                state = "manual"
            elif pid == 254:
                state = "once_program"
            elif pid == 0:
                state = "idle"
            else:
                state = "program"
        else:
            if pid > 0:
                state = "waiting"
            else:
                state = "idle"

        return state
```

```
##### starting station 0 info dump
name: Pump
index: 0
is runnng: True
running program id: 0
seconds remaining: 0
start time: 0
max name length: 32
master 1 operation enabled: False
master 2 operation enabled: False
rain delay ignored: False
sensor 1 ignored: False
sensor 2 ignored: False
enabled: True
sequential operation: False
station type: standard
status: idle
```

It doesn't feel right to be flagged as `idle` when it's on..

vinteo:
I don't have a master so I can't run those scenarios but that makes sense

travisghansen:
What value to make it though..

travisghansen:
Do you think there's any confusion about the switches with `programs` being enabed/disabled vs *running*?

vinteo:
Maybe but now its consistent, switched enable/disable and services to run the actual program/station

travisghansen:
Ah, good point. I'm trying to find some decent docs on how to approach the services. Just reading code ATM but if you've got something to point me at all the better.

Any thoughts on what status value to use for master running?

vinteo:
`master_xxx` maybe? Where it is xxx is program, manual, waiting, once_program?

travisghansen:
Don't really have all that info in that context. Maybe we just set it to `master`. Is there any other scenario where `is_running` is `True` and `pid=0`?

vinteo:
I am not sure, I don't think so

travisghansen:
K, I'm adding `switches` for stations now.

travisghansen:
Shall we add `binary_sensors` for `is_running` for both `stations` and `programs`?

vinteo:
I was just thinking the same thing! lol

travisghansen:
:D Great minds! Would be crucial for automation/rules.

travisghansen:
Got this removing the old `operation` entity:

```
/config/custom_components/opensprinkler/__init__.py:138: RuntimeWarning: coroutine 'Entity.async_will_remove_from_hass' was never awaited
  super().async_will_remove_from_hass()
```

vinteo:
Added an `await` in https://github.com/vinteo/hass-opensprinkler/pull/66

travisghansen:
Looking good!

Screenshot from 2020-05-27 01-02-16

vinteo:
Great, it does feel more stable now. Dinner break for me now, do you want to do the binary sensors? Or I can do them later

travisghansen:
I'm trying to do the `binary_sensor` stuff but the state is hass shows `unknown` for some reason. Trying to figure out what I've got wrong..

vinteo:
You should use the `OpenSprinklerBinarySensor` class like the switches

You can open a PR and what you have and I'll take a look later if you haven't figured it out.

travisghansen:
Yeah, that's exactly what I'm using..strange. I'll poke at it for a few more minutes and then probably get some sleep :)

travisghansen:
Great progress! Time for sleepy sleep for me :)

travisghansen:
This was good to see:

```
2020-05-27 09:03:52 INFO (SyncWorker_25) [backoff] Backing off request_http(...) for 0.8s (pyopensprinkler.OpenSprinklerConnectionError: Cannot connect to controller)
2020-05-27 09:03:52 INFO (SyncWorker_25) [backoff] Backing off request_http(...) for 1.0s (pyopensprinkler.OpenSprinklerConnectionError: Cannot connect to controller)
2020-05-27 09:03:53 ERROR (SyncWorker_25) [backoff] Giving up request_http(...) after 3 tries (pyopensprinkler.OpenSprinklerConnectionError: Cannot connect to controller)
2020-05-27 09:03:53 ERROR (MainThread) [custom_components.opensprinkler] Unable to connect to OpenSprinkler controller: Cannot connect to controller
2020-05-27 09:03:53 WARNING (MainThread) [homeassistant.config_entries] Config entry for opensprinkler not ready yet. Retrying in 80 seconds.
2020-05-27 09:05:29 INFO (MainThread) [homeassistant.components.binary_sensor] Setting up binary_sensor.opensprinkler
2020-05-27 09:05:29 INFO (SyncWorker_21) [homeassistant.loader] Loaded switch from homeassistant.components.switch
2020-05-27 09:05:29 INFO (SyncWorker_32) [homeassistant.loader] Loaded sensor from homeassistant.components.sensor
2020-05-27 09:05:29 INFO (MainThread) [homeassistant.setup] Setting up switch
2020-05-27 09:05:29 INFO (MainThread) [homeassistant.setup] Setup of domain switch took 0.0 seconds.
2020-05-27 09:05:29 INFO (MainThread) [homeassistant.setup] Setting up sensor
2020-05-27 09:05:29 INFO (MainThread) [homeassistant.setup] Setup of domain sensor took 0.0 seconds.
2020-05-27 09:05:29 INFO (MainThread) [homeassistant.components.switch] Setting up switch.opensprinkler
2020-05-27 09:05:29 INFO (MainThread) [homeassistant.components.sensor] Setting up sensor.opensprinkler
```

Pretty sure I saw a bug about this where the power cuts and opensprinkler took longer to come up than hass. In addition, it appears the availability stuff 'just works' as is.

vinteo:
Yes it was mentioned in the HASS PR this is one of the advantages of config flow. Good to see in action

vinteo:
I think now we have feature parity with the current version. Shall we do a 1.0 release this weekend? Gives us couple of days of testing and weekend will give me time to fix anything that comes up.

The only thing I would want to squeeze in is the simplified config. For the icons I believe we have to get that PR in the HA brands repo to be in the custom components folder instead of the core components (and reopen another PR to move from custom components to core components for the official integration). Not sure what the procedure is.

travisghansen:
Sure, I'd like to see the unique id stuff nailed down a little tighter before hand (I don’t think we should be using entry_id how we are and instead use the mac in it’s place and slugify it). I think we can probably get quite a bit more of the items checked off before then as well.

Regarding icons I’m referring to the entity icons and not the integration icon. I think we can find much better defaults for the switches/sensors/etc than the hass defaults.

I’ll have a bit more time this evening (USA) to probably hack a few more things together including various sensors from the controller etc.

vinteo:
The `entry_id` is the unique id generated from the mac address from the config (at least that's how I think it works) so it kinda already does that

travisghansen:
My understanding is different but I’ll do some research/testing and see how it goes. We should be able to customize a device name for example, delete the integration and then re-add the same device and still have the customization persist right?

vinteo:
Yes that is my understanding if you use the same mac address in the config

travisghansen:
Yeah ok. Well I’ll do some testing and see what I observe.

Any comments or observations on any of the sensors added above?

Do you know if multiple programs can be running simultaneously?

vinteo:
For the sensors I think they should be attributes, there are quite a few sensors already and not sure if many people will use them. If the user need sensors for those they can use the template sensor to create one.

Although I have never tried it, I don't see why you can't run multiple programs at once.

travisghansen:
Got an example of how a template sensor would work in this case?

Do you know if attributes can be used in automations/notifications etc easily?

vinteo:
https://www.home-assistant.io/integrations/template/ The first example is one with using attributes. 

You can use the `state_attr()` method in automations and notifications as well I believe

travisghansen:
Ah, then agreed 100% lets keep the pollution down.

travisghansen:
I’ll see if I can figure out how to get some attrs added..

vinteo:
I did some quick tests. For current implementation

1. Added new integration 
2. Customise name of one of the entities
3. Added integration with same mac address with new name -> Updates current integration with new name but entity custom name is gone
4. Customise name of one of the entities again
5. Delete integration
6. Added integration with same mac address -> entity custom name is gone

Then I ran the same steps with the mac address (hardcoded for testing) in the unique id and got the same results. So I guess its not meant to keep the customisation? I ran both test from a fresh copy of the default HA config

vinteo:
Ran same test with the mac address as the identifier as the device as well and same result

travisghansen:
So should we remove the option/notion of the mac altogether?

vinteo:
I think we can keep it. If/when OpenSprinkler adds the mac address to their API, we can gracefully switch over without any user intervention

travisghansen:
I'm failing to see where the binding happens in the code between `mac` and `unique_id`. Either I'm missing it or there's some black magic.

In any case, I think we still have something off on the approach but not entirely sure what the behavior is supposed to be. For example: https://github.com/home-assistant/core/blob/89fe488b7ccee0cf368e284ae0afb812edb9772d/homeassistant/components/samsungtv/media_player.py#L155

Notice they're setting the identifiers to the `unique_id` of the `entry` and not the `entry_id`. I'm not sure exactly how this is supposed to tie into the `unique_id` of the entities generally but I'm guessing something like this is more appropriate: `slugify(f"{entry.unique_id}_{self._entity_type}_program_running_{self._program.index}")`

Perhaps it's best to reach out to the core hass folks about this one.

In any case, for now I'm moving towards the state attrs and I'll see if I can get something in place with those..

vinteo:
I haven't dived into the source but I think the black magic is set here which creates the entry_id for the integration: https://github.com/vinteo/hass-opensprinkler/pull/52/files#diff-a078ebd9bd4dca604dfabb3522103c2aR49

The example you set they actually use the IP address with this method/function which was what was advised against in the other PR. If they used the uuid instead then they should also be able to use the entry_id is my understanding

I am also basing this on the fact they reviewed that part in the HASS PR and only made me remove the name (which makes sense)

travisghansen:
Ah, I missed that due to the constant :( now that makes more sense.

I think there are many other examples: https://github.com/home-assistant/core/blob/89fe488b7ccee0cf368e284ae0afb812edb9772d/homeassistant/components/brother/sensor.py#L46

I've about got state attrs wrapped up and will commit shortly.

vinteo:
I look through the source code a bit and `entry.unique_id` is just the mac address. `entry.entry_id` is also unique to the integration. But from what I can see it wouldn't make a difference which one is used in terms of functionality as the ids are both unique to the integration. But `entry_id` is an actual guid so I think it works better.

travisghansen:
I think that's kinda the point, the device identifier if possible should be deterministic (ie: the mac..which we happen to be setting as the `unique_id`). So no matter how many times the device is added/removed/discovered it's `unique_id` is consistent. I don't know the *why* behind all this, but pretty sure that's what the hass team is after.

Nick-Adams-AU:
Hey team. I have been following the development...great work!

I know this is a work-in-progress but I wanted to check for a couple if a couple of features will be added:

- Ability to run a program manually...bonus points to allow specifying the duration.
- Sensor 1 / rain sensor exposed as a sensor
- Program / zone last run / next run sensors

Thanks again for your work. I'm happy to test anything.

vinteo:
> Hey team. I have been following the development...great work!
> 
> I know this is a work-in-progress but I wanted to check for a couple if a couple of features will be added:
> 
> * Ability to run a program manually...bonus points to allow specifying the duration.

There will be services to run an existing set program or run individual stations for set times. 

> * Sensor 1 / rain sensor exposed as a sensor

This will be attributes of the controller switch which you can use the template sensor to create your own sensor. 

> * Program / zone last run / next run sensors

Same as above

> 
> Thanks again for your work. I'm happy to test anything.

You can pull this branch and test it if you want, it should be pretty much working. I think there might be some final changes to the sensor 1/rain sensor attributes coming though.

travisghansen:
After digging into the sensor stuff a bit more I think we should make them real sensors in hass. The historical data for the sensors along with the flow rate etc I think is important to be first class items.

vinteo:
A template sensor will have historical data as well if created. Its just a matter of whether the majority of people will need them and having tons of sensors vs letting the minority create them when needed. But I am not opposed to creating them.

travisghansen:
Ah good point. I still think it makes sense to have a few more of these for the special cases/data but certainly not for the items that scale linearly like programs and stations..I’ve not got strong feelings either however.

vinteo:
I tend to feel its better to start with less and see the trend and add when needed rather then try to predict the trend

travisghansen:
Yeah, on the flip side using the new flows it's simple to disable entities as well. It would seem odd to me to not have sensors in hass out of the box for attributes of a device that are literally called sensors :)

Nick-Adams-AU:
Thanks @vinteo! I have pulled the branch and am now having a look through the functionality. I appreciate your effort.

I'd like to provide a +1 for having the rain sensor as a first class sensor. It is a key part of any irrigation system knowing whether the rain sensor will stop or programs.

travisghansen:
@Nick-Adams-AU I don't know that I've seen next-run data in the API as distinct datapoints..do you know anything about it?

If they provide datapoints then I'd add it...if we have to calculate it based on the matrix of configuration options probably not..

vinteo:
Good point about the disable. I wonder if we can add it as a disabled entity by default and whether that we should do that or just add it anyway. Seems like people do want it (I don't have any sensors so I'm coming from a different perspective) so maybe we should add it

vinteo:
I do want to get a release out though, since the basic structure is in place, and follow up with others. If we keep trying to add stuff it will be never ending

travisghansen:
Yeah, based on the comment from `core` I say we either:

- use the `mac` for entity `unqiue_id`s

or

- remove the `mac` config option and deal with it if/when the API supports retrieving it

I'm a fan of the former.

https://github.com/home-assistant/core/pull/34146#issuecomment-636211909
https://developers.home-assistant.io/docs/entity_registry_index/#unique-id-requirements

vinteo:
I'll make that change now

travisghansen:
I would `slugify` them while we're at it..

Nick-Adams-AU:
> @Nick-Adams-AU I don't know that I've seen next-run data in the API as distinct datapoints..do you know anything about it?
> 
> If they provide datapoints then I'd add it...if we have to calculate it based on the matrix of configuration options probably not..

It is available using the "/jp" get program data endpoint. It has a funky data structure but it can be calculated off a single HTTP call with no reference to external configuration options.

The end result is that you could say the next run for zone 3 is at "Tuesday 8am for 20 minutes".

Nick-Adams-AU:
With that most recent change (#74) I'm still seeing my 'sensor.opensprinkler_rain_delay_stop_time' showing as 'Unknown' when it should probably be 'None'. The raw value is zero.

vinteo:
@Nick-Adams-AU `None` shows as `Unknown` in HA.

Nick-Adams-AU:
I think run_seconds for the run_station service also needs to be optional as per the service specification:

https://github.com/vinteo/hass-opensprinkler/blob/e1782b4125ea550c61d946d5c003bb2b84f91cd9/custom_components/opensprinkler/sensor.py#L28

Nick-Adams-AU:
> @Nick-Adams-AU `None` state shows as `Unknown` in HA.

Maybe it should be a literal zero then as the value isn't actually 'Unknown' ?

vinteo:
@Nick-Adams-AU not sure `0` makes sense here since its a `timestamp` sensor. I think None/Unknown is correct here

vinteo:
@Nick-Adams-AU interested in becoming a collaborator? At minimum you could do some reviews/testing :)

Nick-Adams-AU:
Happy to help with whatever I can! I had only found this branch last night but have now upgraded and am running it through its paces. I'll shoot through anything else I find. At any rate, you've already done 95% of the work :)

travisghansen:
I'll run through all the `_time` sensors and return `None` instead of `0` yeah?

vinteo:
@travisghansen yea

travisghansen:
@Nick-Adams-AU that's some pretty complex logic for the start times. Also, it's on the `program` level and not the `station` level...so it would be even more complex to determine the next run for a `program` but seems possible. Given the nasty matrix involved there I'm not gonna tackle that one just yet. We'll focus on getting a release out from where we're at and then can revisit that. Maybe create a separate ticket here or in the other project to address it.

vinteo:
We should also consider providing the datapoints only (in a logical way of course) in the library and the attributes here and letting the user format it the way they want in HA.

travisghansen:
@vinteo yeah agreed. Problem is the raw data results in some pretty nasty logic to figure out what's really going on. `hass` would have a really hard time digesting the info in a remotely sane way without a fairly substantial lift.

Nick-Adams-AU:
I totally agree that it is a complex data structure and probably best left for later on. Exposing it in the library sounds like a good start.

I think we'd only need the program next run time as that is the only thing that actually has schedules.

travisghansen:
@Nick-Adams-AU yeah, I think we can get there but it might take a bit. The different options for weekday/interval + sunset/sunrise etc etc make it pretty nasty :(

vinteo:
OK guys! Is it time for a release? :D

travisghansen:
I'll go through and add all the new properties now and update the `_times` to return `None` if `0`.

After that I'll take a closer look at the services.

vinteo:
I think its time for a feature freeze on this branch, the only thing that should go in from now is:

- the fix for the error thrown for old/unavailable attributes

travisghansen:
Do we want to get the update options flow in there before 1.0?

vinteo:
I was looking at the OptionsFlow and it didn't look like it was for updating the config but add other options to it. Either way I think it would be a too complex feature to add without some proper thought and testing around.

travisghansen:
Ok, I think I saw it in the `nut` integration...maybe I’ll have a look.

vinteo:
Once this goes green I am gonna merge to master and maybe do the release :)

Unless there are any objections @travisghansen

Nick-Adams-AU:
Great work guys!!

travisghansen:
@vinteo I think this is a fantastic 1.0! Thanks for working together to make it happen!

vinteo:
Great work @travisghansen , your motivation really pushed me as well. Thank you very much

vinteo:
Release done *pops champagne* 🍾

travisghansen:
@vinteo some day I'll go to Australia and buy you a beer :D

vinteo:
@travisghansen if you are coming to Australia, the beer's on me :D

LordShadowen:
hi guys - not sure if this is right place to comment or if I should open an issue (apologies if I got it wrong...)

I got a notification that issue #35 (Include "remaining water time" in sensors) was getting addressed by release 1.0 which got me all excited to try it out (and I just started doing...)

I'm personally not a fan of having UI-only integrations in HA. I understand that it seems to be the "standard" in the project going forward, and I assume that's why there is no longer a YAML configuration for this component. 

Again, not happy about it, but fair enough - I can live with it :)
But providing the password in the UI text and having it stored clear text in the core.config_entries file is a much bigger no-no. How is everyone handling that? 
Is there a way to use the secrets file in the UI configuration?

It's probably a bigger discussion in HA than in this one component, but I was curious what you're thoughts are.

travisghansen:
@LordShadowen thanks for reaching out! If you want more info probably open a new issue to discuss but I'll shed some light here real quick:

- We're trying to get the project into hass core and they basically said "don't do yaml configuration"..so is what it is :(
- The password in the UI stored in plain text is no better or worse than the md5 hash is/was. Technically the hash is what's used for API communication so in that sense it *is* the pw for the API. As an example the hash could be used (without any other info) to disable passwords altogether

You're right that it's probably a bigger discussion with the larger hass ecosystem. I'm new to hass but I assume eventually they'll introduce some form of encryption key that will be used to decrypt values stored in yaml/etc.

Looking forward to feedback on the remaining water time and all the other new features!

LordShadowen:
> * We're trying to get the project into hass core and they basically said "don't do yaml configuration"..so is what it is :(

Right, that's what I figured :)

> * The password in the UI stored in plain text is no better or worse than the md5 hash is/was. Technically the hash is what's used for API communication so in that sense it _is_ the pw for the API. As an example the hash could be used (without any other info) to disable passwords altogether

I'm not questioning using the OpenSprinkler password vs the hash. That's fine. It's just the storing of either of those in plain text in a non-centralized, special file that we don't manually control. 
When using Yaml, the password could be kept in the secrets file and that allowed for a neat separation and centralization (everything that is "secret" is there). 
I wish that at very least HA was allowing a special syntax of "!secret XXX" in UI configuration fields or something similar.

But yes, that's a discussion for another day and place. Sorry for the small rant, not your fault ;) 
I'll check out the actual component later this week and will let you know if I spot any problems. 
Cheers

PR #73, opened 2020-05-28
state attributes for controller, stations, and programs

travisghansen:


custom_components/opensprinkler/__init__.py
```diff
@@ -115,6 +115,117 @@ def _get_state(self):
         """Retrieve the state."""
         raise NotImplementedError
 
+    def _get_controller_attributes(self, controller):
+        attributes = {}
+        return attributes
+
+    def _get_program_attributes(self, program):
+        attributes = {}
+        return attributes
```
vinteo, line 124: supposed to remove these?

travisghansen: lol good call :) removed

Issue #72, opened 2020-05-28
Icons Audit


vinteo:
Icons: https://cdn.materialdesignicons.com/4.5.95/

Controller Switch: ?? (or keep default lightning)
Water Level Sensor: mdi:water -> mdi:water-percent
Last Run Sensor: mdi:history (keep)
Rain Delay Stop Time: mdi:weather-rainy (keep)
Program Binary Sensor: mdi:calendar-clock
Program Switch: mdi:calendar-clock (or keep default lightning)
Station Sensor: -> mdi:water-pump, mdi:water, mdi:sprinkler, mdi:sprinkler-variant
Station Binary Sensor: -> mdi:water-pump, mdi:water, mdi:sprinkler, mdi:sprinkler-variant
Station Switch: -> mdi:water-pump, mdi:water, mdi:sprinkler, mdi:sprinkler-variant (or keep default lightning)

vinteo:
@travisghansen my suggestions!

travisghansen:
I'll check them out and see if I have other suggestions.

travisghansen:
Just having fun..thoughts?

```
diff --git a/custom_components/opensprinkler/binary_sensor.py b/custom_components/opensprinkler/binary_sensor.py
index 75581a5..790b9c1 100755
--- a/custom_components/opensprinkler/binary_sensor.py
+++ b/custom_components/opensprinkler/binary_sensor.py
@@ -79,6 +79,14 @@ class ProgramIsRunningBinarySensor(OpenSprinklerBinarySensor, BinarySensorEntity
         """Return a unique, Home Assistant friendly identifier for this entity."""
         return f"{self._entry_id}_{self._entity_type}_program_running_{self._program.index}"
 
+    @property
+    def icon(self) -> str:
+        """Return icon."""
+        if self._program.is_running:
+            return "mdi:timer"
+        
+        return "mdi:timer-off"
+
     def _get_state(self) -> bool:
         """Retrieve latest state."""
         return bool(self._program.is_running)
@@ -107,6 +115,14 @@ class StationIsRunningBinarySensor(OpenSprinklerBinarySensor, BinarySensorEntity
         """Return a unique, Home Assistant friendly identifier for this entity."""
         return f"{self._entry_id}_{self._entity_type}_station_running_{self._station.index}"
 
+    @property
+    def icon(self) -> str:
+        """Return icon."""
+        if self._station.is_running:
+            return "mdi:water"
+        
[...]

PR #74, opened 2020-05-28
icons, minor state attrs adjustments, None when 0 for timestamp

travisghansen:
Initial stab at icons.

Closes: #72

### DIFF
--- a/custom_components/opensprinkler/switch.py
+++ b/custom_components/opensprinkler/switch.py
@@ -71,12 +71,20 @@ class ControllerOperationSwitch(OpenSprinklerBinarySensor, SwitchEntity):
     @property
     def name(self):
         """Return the name of controller switch."""
-        return f"{self._name} Operation Enabled"
+        return f"{self._name} Enabled"
 
     @property
     def unique_id(self) -> str:
         """Return a unique, Home Assistant friendly identifier for this entity."""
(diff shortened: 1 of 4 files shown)


custom_components/opensprinkler/switch.py
```diff
@@ -71,13 +71,21 @@ def __init__(self, entry_id, name, controller, coordinator):
...
     @property
     def unique_id(self) -> str:
         """Return a unique, Home Assistant friendly identifier for this entity."""
         return f"{self._entry_id}_{self._entity_type}_controller_operation_enabled"
 
+    @property
+    def icon(self) -> str:
+        """Return icon."""
+        if self._controller.operation_enabled:
+            return "mdi:circle"
```
vinteo, line 85: Leave as lightning

travisghansen: @vinteo any interest in the `power-*` icons?

vinteo: how about something a bit more out there... `barley` and `barley-off`

vinteo: or `leak` and `leak-off`

travisghansen: I like the `barley` better than the `leak` :D

vinteo: Let's change it barley and merge this

PR #76, opened 2020-05-29
refactor: use entry.unique_id in entity unique_id

vinteo:


custom_components/opensprinkler/__init__.py
```diff
@@ -222,7 +222,7 @@ def _get_station_attributes(self, station):
     def device_info(self):
         """Return device information about Opensprinkler Controller."""
         return {
-            "identifiers": {(DOMAIN, self._entry_id)},
+            "identifiers": {(DOMAIN, self._entry.unique_id)},
```
vinteo, line 225: do I need to slugify here?

vinteo: added it

travisghansen:
@vinteo should we also `slugify` the value here? `await self.async_set_unique_id(user_input[CONF_MAC])`

vinteo:
maybe it would be a good idea, i think it will convert to lower case as well so it would account for user variation in that

travisghansen:
Yeah, let's do it. Then I'll merge.

vinteo:
done

PR #79, opened 2020-05-30
fix: make run_seconds in service optional

vinteo:


Nick-Adams-AU:
I think this still needs a default run_seconds as pyopensprinkler won't provide one.

travisghansen:
@Nick-Adams-AU `pyopensprinkler` defaults to 60 seconds.

vinteo:
Here: https://github.com/vinteo/py-opensprinkler/blob/master/pyopensprinkler/station.py#L65

I tested it so it does work

travisghansen:
@vinteo FYI I just hacked that method to default to `None` to prevent the crazy python behavior as an FYI.

PR #80, opened 2020-05-30
chore: update to pyopenpsprinkler 0.6.0

vinteo:


travisghansen:
@vinteo You want to wait and pull in the `run_once_program`?

vinteo:
yea since that is ready, just a sec

vinteo:
I updated the version but I think the service will wait for the next release.

PR #81, opened 2020-05-30
refactor: simplify services

vinteo:


travisghansen:
Still haven't delved into the code but wanted to mention this one `controller.stop_all_stations` as a candidate.

Nick-Adams-AU:
`controller.stop_all_stations` is a great idea!

vinteo:
https://github.com/vinteo/hass-opensprinkler/issues/84

vinteo:
this is ready to go if ok

travisghansen:
@vinteo I think the default values need to be set to `None` and do the whole `if run_seconds is None` business.

The risk we run as is, if something invokes the method with a a non-default value, anything subsequently invoking the method without a default will inherit effectively whatever last changed the value. It's craziness I know.

vinteo:
I removed the defaults instead, and it should be handled by the library now?

travisghansen:
@vinteo I believe you are correct actually. Which is great. LGTM

travisghansen:
@vinteo actually, one last bit, does it make any sense to unify the methods under `OpenSprinklerEntity`? Implement it once everywhere?

vinteo:
I used class inheritance, I think i got it right...

custom_components/opensprinkler/__init__.py
```diff
@@ -293,3 +293,19 @@ class OpenSprinklerSensor(OpenSprinklerEntity):
...
+
+
+class OpenSprinklerProgramEntity:
+    def run(self):
+        """Runs the program."""
+        return self._program.run()
+
+
+class OpenSprinklerStationEntity:
+    def run(self, run_seconds=None):
```
vinteo, line 305: needed to add the `None` for optional

travisghansen: Looks fantastic.

Issue #84, opened 2020-05-30
Add service to stop all stations


vinteo:
run on controller entities

travisghansen:
For ease of use does it make any sense to make it available to any entity?

vinteo:
definitely makes it easier but not sure we should

travisghansen:
Agreed.

travisghansen:
Do the method names on the objects have to match exactly what the service name is called? I've not wrapped my brain around the binding here yet..

vinteo:
the method name is the third param (string) of register service method. I think should be `stop_all` for this case

travisghansen:
Ok cool. Go ahead on this one similar to the other I suppose using the inheritance stuff.

PR #86, opened 2020-05-30
feat: stop all station service

vinteo:
Closes: #84

### DIFF
--- a/custom_components/opensprinkler/__init__.py
+++ b/custom_components/opensprinkler/__init__.py
@@ -367,6 +367,12 @@ class OpenSprinklerSensor(OpenSprinklerEntity):
         return self._state
 
 
+class OpenSprinklerControllerEntity:
+    def stop(self):
+        """Stops all stations."""
+        return self._controller.stop_all_stations()
+
+
 class OpenSprinklerProgramEntity:
     def run(self):
         """Runs the program."""
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Note: *1.0.0 has major breaking changes, you will need to update any automations
 
 ## Using Services
 
-Available services are `opensprinkler.run` for programs and stations, and `opensprinkler.stop` for stations.
+Available services are `opensprinkler.run` for programs and stations, and `opensprinkler.stop` for stations or controller (to stop all stations).
 
 ```yaml
 service: opensprinkler.run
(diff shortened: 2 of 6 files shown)


custom_components/opensprinkler/__init__.py
```diff
@@ -367,6 +367,14 @@ def state(self):
         return self._state
 
 
+class OpenSprinklerControllerEntity:
+    def stop(self):
+        """Stops all stations."""
+        for _, station in self._controller.stations.items():
```
travisghansen, line 373: I think we should use this instead: `controller.stop_all_stations()` Although maybe we should rename that method and/or the service. Basically it stops any running stations *and* any stations that are `waiting`.

custom_components/opensprinkler/__init__.py
```diff
@@ -367,6 +367,14 @@ def state(self):
         return self._state
 
 
+class OpenSprinklerControllerEntity:
+    def stop(self):
```
travisghansen, line 371: I'd probably name this a little more explicitly. `stop_stations` and then if we implement the other method `reset_stations`?

travisghansen: I've just updated the lib to have both methods. I think we should have both services.

vinteo: I didn't realise there was that method. Is there a scenario where we would want to stop all stations but not the waiting? Maybe the reset should be enough?

travisghansen: Yeah, good question. I've been surprised by the general sense that folks are wanting to build some pretty power scheduling/etc logic outside the unit itself (I may be misreading the sentiment). My comment is mostly driven by that feeling.

In that same vein I'd suggest we also add `enable`/`disable` services for the controller and possible even `reboot`.

vinteo: The controller already has an enable/disable switch, I don't see why it needs a service. The reboot service we can add if it is requested, but I don't see the need to add things that potentially might never get used.

vinteo: For this PR I think we should just keep the one service

travisghansen: That's fair...haven't delved into the automations too deeply yet so I'm a bit blind on this stuff. Meaning, if switches can generally be toggled in automations and don't require services then yeah, agreed.

custom_components/opensprinkler/__init__.py
```diff
@@ -367,6 +367,12 @@ def state(self):
...
+        """Stops all stations."""
+        return self._controller.stop_all_stations()
```
vinteo, line 373: @travisghansen changed to just call the "reset"

PR #96, opened 2020-06-02
fix: pyopensprinkler 0.6.4, minor fixes

vinteo:


custom_components/opensprinkler/__init__.py
```diff
@@ -200,13 +200,12 @@ def _get_controller_attributes(self, controller):
             "last_reboot_time",
         ]:
             iso_attr = attr + "_iso"
-            if getattr(controller, attr) == 0:
+            timestamp = getattr(controller, attr)
+            if not timestamp:
```
vinteo, line 204: this accounts for a new controller that has never been run so timestamp is `None` as well as 0

travisghansen:
LGTM

Issue #97, opened 2020-06-03
SCAN_INTERVAL not used for updates


JurajNyiri:
SCAN_INTERVAL is currently not used for updates and it takes up to 30 seconds for a component to update.

- SCAN_INTERVAL **does** work in the refresh function in init  _async_update_listener_action
- SCAN_INTERVAL **does not** work in update function in init

That means, that internally the data are refreshed every SCAN_INTERVAL seconds, but are reported to HASS only every 30 seconds.

Replication steps:

Modify update function in __init__.py to this:
```
@Throttle(SCAN_INTERVAL)
    def update(self) -> None:
        _LOGGER.warn("TEST2")
        """Update latest state."""
        self._state = self._get_state()
```

Watch logs... new log appears only once every 30 seconds.

SCAN_INTERVAL **does** work in the refresh function:

```
async def _async_update_listener_action(self, now):
        _LOGGER.warn("TEST3")
        """Define an async_track_time_interval action to update data."""
        await self._hass.async_add_executor_job(self._controller.refresh)
```

~~Maybe `async_update` should be used instead?~~ (Edit: no, that doesn't run at all)

JurajNyiri:
Managed to fix it, will create PR soon.

PR #98, opened 2020-06-03
fix: use scan_interval correctly

vinteo:
Fixes: #97

### DIFF
--- a/custom_components/opensprinkler/__init__.py
+++ b/custom_components/opensprinkler/__init__.py
@@ -1,6 +1,5 @@
 """OpenSprinkler integration."""
 import asyncio
-from datetime import timedelta
 import logging
 
 from pyopensprinkler import (
@@ -341,7 +340,6 @@ class OpenSprinklerEntity(RestoreEntity):
         await super().async_will_remove_from_hass()
         self._coordinator.deregister_time_interval_listener()
 
-    @Throttle(SCAN_INTERVAL)
     def update(self) -> None:
         """Update latest state."""
         self._state = self._get_state()
--- a/custom_components/opensprinkler/binary_sensor.py
+++ b/custom_components/opensprinkler/binary_sensor.py
@@ -21,6 +21,7 @@ from .const import (
     DOMAIN,
     SERVICE_RUN,
     SERVICE_STOP,
+    SCAN_INTERVAL,
 )
 
 _LOGGER = logging.getLogger(__name__)
--- a/custom_components/opensprinkler/sensor.py
+++ b/custom_components/opensprinkler/sensor.py
@@ -16,7 +16,7 @@ from . import (
     OpenSprinklerSensor,
     OpenSprinklerStationEntity,
 )
-from .const import CONF_RUN_SECONDS, DOMAIN, SERVICE_RUN, SERVICE_STOP
+from .const import CONF_RUN_SECONDS, DOMAIN, SERVICE_RUN, SERVICE_STOP, SCAN_INTERVAL
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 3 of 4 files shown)


custom_components/opensprinkler/__init__.py
```diff
@@ -341,7 +340,6 @@ async def async_will_remove_from_hass(self):
         await super().async_will_remove_from_hass()
         self._coordinator.deregister_time_interval_listener()
 
-    @Throttle(SCAN_INTERVAL)
```
JurajNyiri, line 344: @Throttle didn't cause issues in my testing and I use it in my custom components, what is the reason for removal?

custom_components/opensprinkler/__init__.py
```diff
@@ -19,7 +18,7 @@
 from homeassistant.util import slugify, Throttle
 from homeassistant.util.dt import utc_from_timestamp
 
-from .const import DEFAULT_PORT, DOMAIN, SCAN_INTERVAL
+from .const import DEFAULT_PORT, DOMAIN, UPDATE_INTERVAL
```
JurajNyiri, line 21: I would personally leave the original var name SCAN_INTERVAL, it is a standard across hass components

vinteo: I am not sure how python works but I wanted to keep it in one place so can import it, it is set as `SCAN_INTERVAL` below

custom_components/opensprinkler/__init__.py
vinteo: Throttle from my understanding is to stop it going too fast, for example limiting updates that call API too fast, so if it is < 30 secs it doesn't do anything

vinteo: changed it slightly

custom_components/opensprinkler/__init__.py
vinteo: changed it slightly

custom_components/opensprinkler/binary_sensor.py
```diff
@@ -21,9 +21,11 @@
     DOMAIN,
     SERVICE_RUN,
     SERVICE_STOP,
+    UPDATE_INTERVAL,
```
JurajNyiri, line 24: UPDATE_INTERVAL as SCAN_INTERVAL

custom_components/opensprinkler/sensor.py
```diff
@@ -16,9 +16,10 @@
     OpenSprinklerSensor,
     OpenSprinklerStationEntity,
 )
-from .const import CONF_RUN_SECONDS, DOMAIN, SERVICE_RUN, SERVICE_STOP
+from .const import CONF_RUN_SECONDS, DOMAIN, SERVICE_RUN, SERVICE_STOP, UPDATE_INTERVAL
```
JurajNyiri, line 19: UPDATE_INTERVAL as SCAN_INTERVAL

custom_components/opensprinkler/switch.py
```diff
@@ -14,7 +14,9 @@
     OpenSprinklerProgramEntity,
     OpenSprinklerStationEntity,
 )
-from .const import CONF_RUN_SECONDS, DOMAIN, SERVICE_RUN, SERVICE_STOP
+from .const import CONF_RUN_SECONDS, DOMAIN, SERVICE_RUN, SERVICE_STOP, UPDATE_INTERVAL
```
JurajNyiri, line 17: UPDATE_INTERVAL as SCAN_INTERVAL

custom_components/opensprinkler/__init__.py
JurajNyiri: You could import it as SCAN_INTERVAL, that worked in my fix.. if you want to keep name UPDATE_INTERVAL see my recommendation below

vinteo: nah it that works I will change it all to SCAN_INTERVAL. Not very experienced with python so bear with me :)

vinteo:
Duplicate of #99

PR #99, opened 2020-06-03
Fix: SCAN_INTERVAL not used for updates

JurajNyiri:
Fixes: #97


JurajNyiri:
Feel free to close/merge, this fixes the same thing as https://github.com/vinteo/hass-opensprinkler/pull/98 without the new variable... as discussed on slack.

custom_components/opensprinkler/__init__.py
```diff
@@ -341,7 +341,6 @@ async def async_will_remove_from_hass(self):
         await super().async_will_remove_from_hass()
         self._coordinator.deregister_time_interval_listener()
 
-    @Throttle(SCAN_INTERVAL)
```
vinteo, line 344: Can yo also remove the timedelta import since its not used. and also the file permissions for this file was changed, you might want to check your IDE

JurajNyiri: 👍 Addressed both, also removed Throttle from imports

vinteo:
Thanks for your contribution

PR #106, opened 2020-06-11
async/coordinator support

travisghansen:
Use the upstream `coordinator` and move to `async` as a result. Would be ideal to make a new snap of the library with the `lock` in place and include the update library here.

### DIFF
--- a/custom_components/opensprinkler/__init__.py
+++ b/custom_components/opensprinkler/__init__.py
@@ -375,7 +375,7 @@ class OpenSprinklerSensor(OpenSprinklerEntity):
 
 
 class OpenSprinklerControllerEntity:
-    def run(self, run_seconds=None):
+    async def run(self, run_seconds=None):
         """Run once program."""
         if run_seconds == None or (
             not isinstance(run_seconds, list) and not isinstance(run_seconds, dict)
@@ -392,7 +392,11 @@ class OpenSprinklerControllerEntity:
                     if run_seconds.get(str(x)) is not None
                     else 0
                 )
-            return self._controller.run_once_program(run_seconds_list)
+            await self.hass.async_add_executor_job(
+                self._controller.run_once_program, run_seconds_list
+            )
+            await self._coordinator.async_request_refresh()
+            return
 
         if not isinstance(run_seconds[0], int):
             run_seconds_by_index = {}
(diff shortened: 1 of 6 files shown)


custom_components/opensprinkler/__init__.py
```diff
@@ -38,8 +49,30 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
...
+            """Fetch data from OpenSprinkler."""
+            print("refreshing OpenSprinkler data")
+            async with async_timeout.timeout(TIMEOUT):
+                await hass.async_add_executor_job(controller.refresh)
+                if not controller._state:
+                    raise UpdateFailed("Error fetching OpenSprinkler state")
+
+                return controller._state
+
+        scan_interval = entry.data.get(CONF_SCAN_INTERVAL, DEFAULT_SCAN_INTERVAL)
```
vinteo, line 64: I guess this just get the default ATM? Just FYI, this conf will need to go into `OptionFlow` rather than `ConfigFlow` to go into core. So a refactor will be needed later

vinteo:
Can you also update to the latest `pyopensprinkler` release

custom_components/opensprinkler/switch.py
```diff
@@ -90,15 +89,15 @@ def _get_state(self) -> str:
         """Retrieve latest state."""
         return bool(self._controller.enabled)
 
-    def turn_on(self, **kwargs):
+    async def async_turn_on(self, **kwargs):
```
vinteo, line 92: I am thinking we may also need to update the service `run` and `stop` to async

travisghansen:
Yeah did we get a new snap?

custom_components/opensprinkler/switch.py
travisghansen: Yeah I wasn’t sure what was needed for that so I left them. Want me to make the changes or you want to push a commit to this branch?

custom_components/opensprinkler/__init__.py
travisghansen: Ah OK. So what changes do I need to make here?

vinteo: To be honest I am not sure as I have never done `OptionFlow` before but this was a comment on the core PR as I initially had a run seconds config and I assume this will apply for the scan interval as well. 

But I don't think it is a small change so its probably worth to get this in first.

custom_components/opensprinkler/switch.py
vinteo: If it still works maybe we can do it separately in another PR

vinteo: But if you want to do it then go ahead,

custom_components/opensprinkler/__init__.py
travisghansen: I think it may be `entry.options` is all versus `entry.data`. Want me to tweak that?

vinteo: I don't think its that simple as `entry.options` will be empty here, unless options `OptionFlow` goes through the whole setup flow again. Since it is an unknown let's keep it as it is and do it properly when you do the `OptionFlow`

travisghansen: Well, the values in `entry.options` come from the options flow yes. I'm not skilled enough to do that but at least it's looking in the right place now.

vinteo:
Let me know if it is ready for merge

custom_components/opensprinkler/switch.py
travisghansen: I can just try changing them to async and see what happens.

travisghansen:
Looks to be exactly what they want regarding `options`.

https://github.com/home-assistant/core/blob/dev/homeassistant/components/nut/__init__.py#L57

custom_components/opensprinkler/switch.py
travisghansen: Do you know how to pass arguments to this kind of invocation?

`hass.async_add_executor_job(hub.update)`

vinteo: I believe its: 

`hass.async_add_executor_job(hub.update, arg1, arg2, ...)`

travisghansen: OK, I think that worked. If you want to try out the branch and see how it goes for you that would be great.  Didn't test the service calls extensively but I believe the syntax is correct.

travisghansen:
Ready when you are with this one.

vinteo:
Did a quick test on the services, looks to be working.

Merging...

Issue #104, opened 2020-06-09
Run Once Program turns off all the other programs


JurajNyiri:
When using a new run once program feature, to turn on multiple stations at the same time, it turns off all the others.

Replication steps:
1. Turn on s08
2. Run this:
```
- service: opensprinkler.run
   data:
        entity_id: switch.opensprinkler_enabled
        run_seconds:
          - index: 0
            run_seconds: 64800
          - index: 2
            run_seconds: 64800
          - index: 5
            run_seconds: 64800
```
3. s08 is now turned off

Now this might be an intended behaviour but it surprised me. I have many automations turning on/off stations independently from each-other and since implementing the run-once-programs new feature, it has been turning all the others off.

Description should be updated with a **bold** warning or if possible, this behaviour should be changed.

For now I am going back to sequential turn/on/off of individual stations.

vinteo:
Yea this is the side effect of the OS API as it only accepts the run time in the form of an array so it is needed to pad the other stations with 0 seconds.

I will make it clearer in the README etc.

I haven't tested it but does running preset programs in OS also stop running stations? Maybe that's why its called running once program rather than start stations

JurajNyiri:
@vinteo thank you for the explanation!

New feature request:

Could we in the future keep the end times of all the stations and when a new "multi-start" request is sent for some stations, calculate and automatically put in correct values for currently turned on stations instead of 0?

travisghansen:
Interesting idea!

vinteo:
@JurajNyiri please take a look at https://github.com/vinteo/hass-opensprinkler/pull/108

PR #108, opened 2020-06-12
feat: continue_running_stations flag for run once program service

vinteo:
Fixes: #104

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,32 @@ data:
     2: 30
 ```
 
+By default running once program will stop all other stations that are running, you can specify
+`continue_running_stations` to true to allow the stations to continue running. This only works when
+specifying the run seconds in index/second pairs.
+
+```yaml
+service: opensprinkler.run
+data:
+  entity_id: switch.controller_name # Switches or sensors for controller
+  run_seconds: # List of station index and run seconds pairs (required)
+    - index: 0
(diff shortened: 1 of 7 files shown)


vinteo:
@travisghansen flipped the flag around

travisghansen:
Is there a reason this is only supported with the key/value syntax?

vinteo:
You have to pad it in an array so a 0 is given. I don't think it's right to
over write that but that's just me

travisghansen:
Not sure I follow that last comment. I understand the padding the array but what’s getting overwritten?

vinteo:
for example if you want to run station 1 and 3 you have to provide [0, 60, 0, 60], I don't think its right to overwrite the `0` of stations 0 and 2, even with the flag. If the key/value way they still provide a 0 to station 2 to force stop it with the flag and keep station running. 

For example is 0 and 2 is running, they can do
```yaml
run_seconds:
  1: 60
  2: 0
  3: 30
continue_running_stations: True
```
to keep station 0 running but force station 2 to stop.

travisghansen:
Oh right. I was thinking about doing run once on the station for some reason and got my wires crossed. I agree with your assessment, merge away!

This is a great feature add for sure.

vinteo:
Merging...

PR #109, opened 2020-06-16
fix: gracefully fail for deleted program

vinteo:


custom_components/opensprinkler/__init__.py
```diff
@@ -293,11 +293,25 @@ def device_state_attributes(self):
...
+                return False
+
         attributes["program_enabled_count"] = len(
-            dict(filter(lambda e: e[1].enabled == True, controller.programs.items()))
+            dict(filter(lambda e: is_enabled(e[1]), controller.programs.items()))
         )
+
+        def is_running(program):
+            try:
+                return program.enabled == True
```
travisghansen, line 309: Wrong property?

custom_components/opensprinkler/__init__.py
```diff
@@ -361,6 +375,13 @@ async def stop(self):
 
 
 class OpenSprinklerProgramEntity:
+    @property
+    def available(self):
```
travisghansen, line 379: This should ultimately call super I think. Meaning, fall back to the coordinator value.

custom_components/opensprinkler/switch.py
```diff
@@ -124,14 +125,20 @@ def unique_id(self) -> str:
     @property
     def icon(self) -> str:
         """Return icon."""
-        if self._program.enabled:
-            return "mdi:calendar-clock"
+        try:
+            if self._program.enabled:
```
travisghansen, line 129: Did you want to use the new function here?

custom_components/opensprinkler/switch.py
```diff
@@ -124,14 +125,20 @@ def unique_id(self) -> str:
...
+        except:
+            pass
 
         return "mdi:calendar-remove"
 
     def _get_state(self) -> str:
         """Retrieve latest state."""
-        return bool(self._program.enabled)
+        try:
+            return bool(self._program.enabled)
```
travisghansen, line 139: Did you want to use the new function here?

vinteo:
Yea I not that happy with this even though I went through so many iterations while working on this. Python seems to have a pattern of "do then ask for forgiveness" using try...except which I find a little strange lol

travisghansen:
Yeah I don’t know that I’ll come up with anything myself...but I’ll sit down and try something and see what comes out.

travisghansen:
I’ve been thinking about this one a bit and think putting a callback system in place in the library for programs being added/removed may simplify a bit. Before messing with it I thought I’d get some thoughts on the approach?

Then we can register and deregister the entities as appropriate without weird stuff going on..

vinteo:
I think the idea is OK but it might be over engineering it a bit as the outcome will still probably be the same due to the potential index changing for programs in this scenario. And this is not really a common case which can be solved by a restart. 

However, anything is better than nothing that we have at the moment, as long as its not too complicated. But what will trigger the callback though? Seems like there still needs to be some sort of polling unless you mean trigger when there is an error caught? I think will be difficult with the indexes, but maybe you have a solution in mind.

travisghansen:
The library would trigger based on the polling. I’m thinking have some sort of hash based on id/name and if they differ fire the appropriate events..

The indices may change but with hass removing and readding it should be transparent to end user. Especially if we implement deterministic entity ids that don’t change with the name(s).

vinteo:
The index issue weird cause it is the really only "unique" identifier for the program (name maybe but that can change too) but when it changes it is not replaced with another "unique" identifier but potentially given one that references a previous program. I can't really imagine how it would work but maybe I just can't see it right now, probably need to take a look at the solution you have in mind when you get to it.

PR #111, opened 2020-06-20
[WIP] feat: mqtt support

vinteo:
~~Proof of concept, will need much more refactor, requires pyopenspinkler library from https://github.com/vinteo/py-opensprinkler/pull/39~~

### DIFF
--- a/custom_components/opensprinkler/__init__.py
+++ b/custom_components/opensprinkler/__init__.py
@@ -130,10 +130,8 @@ async def async_unload_entry(hass: HomeAssistant, entry: ConfigEntry):
 class OpenSprinklerEntity(RestoreEntity):
     """Define a generic OpenSprinkler entity."""
 
-    def __init__(self, entry, name, coordinator, controller):
+    def __init__(self, entry, name):
         """Initialize."""
-        self._coordinator = coordinator
-        self._controller = controller
         self._entry = entry
         self._name = name
 
@@ -141,12 +139,17 @@ class OpenSprinklerEntity(RestoreEntity):
         """Retrieve the state."""
         raise NotImplementedError
 
+    def _get_controller(self):
+        return self.hass.data[DOMAIN][self._entry.entry_id]["controller"]
+
+    def _get_coordinator(self):
+        return self.hass.data[DOMAIN][self._entry.entry_id]["coordinator"]
+
     @property
     def device_info(self):
         """Return device information about Opensprinkler Controller."""
 
-        controller = self.hass.data[DOMAIN][self._entry.entry_id]["controller"]
-
+        controller = self._get_controller()
         return {
             "identifiers": {(DOMAIN, slugify(self._entry.unique_id))},
             "name": self._name,
@@ -163,16 +166,16 @@ class OpenSprinklerEntity(RestoreEntity):
     @property
     def available(self):
         """Return if entity is available."""
-        return self._coordinator.last_update_success
+        return self._get_coordinator().last_update_success
 
     async def async_added_to_hass(self):
         self.async_on_remove(
-            self._coordinator.async_add_listener(self.async_write_ha_state)
+            self._get_coordinator().async_add_listener(self.async_write_ha_state)
         )
 
     async def async_update(self):
         """Update latest state."""
-        await self.hass.async_add_executor_job(self._controller.refresh)
+        await self.hass.async_add_executor_job(self._get_controller().refresh)
         self.async_write_ha_state()
 
 
@@ -211,7 +214,7 @@ class OpenSprinklerControllerEntity:
 
         if isinstance(run_seconds, dict):
             run_seconds_list = []
-            for _, station in self._controller.stations.items():
+            for _, station in self._get_controller().stations.items():
                 run_seconds_list.append(
                     run_seconds.get(str(station.index))
                     if run_seconds.get(str(station.index)) is not None
@@ -222,9 +225,9 @@ class OpenSprinklerControllerEntity:
                     )
                 )
             await self.hass.async_add_executor_job(
-                self._controller.run_once_program, run_seconds_list
+                self._get_controller().run_once_program, run_seconds_list
             )
-            await self._coordinator.async_request_refresh()
+            await self.async_update()
             return
 
         if not isinstance(run_seconds[0], int):
@@ -235,7 +238,7 @@ class OpenSprinklerControllerEntity:
                 ] = run_seconds_config[CONF_RUN_SECONDS]
 
             run_seconds_list = []
-            for _, station in self._controller.stations.items():
+            for _, station in self._get_controller().stations.items():
                 run_seconds_list.append(
                     run_seconds_by_index.get(station.index)
                     if run_seconds_by_index.get(station.index) is not None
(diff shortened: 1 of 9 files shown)


custom_components/opensprinkler/__init__.py
```diff
@@ -47,12 +52,32 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
...
+        controller = OpenSprinkler(url, password, opts)
         controller.refresh_on_update = False
 
+        if mqtt_host:
+            await hass.async_add_executor_job(controller.mqtt_start)
+
         async def async_update_data():
             """Fetch data from OpenSprinkler."""
+
+            if mqtt_host:  # skip refresh
```
vinteo, line 78: I did a quick and dirty skip here, we can probably switch the whole coordinator out if MQTT is setup

custom_components/opensprinkler/config_flow.py
```diff
@@ -27,6 +36,10 @@
         vol.Required(CONF_PASSWORD): str,
         vol.Required(CONF_MAC): str,
         vol.Optional(CONF_NAME, default=DEFAULT_NAME): str,
+        vol.Optional(CONF_MQTT_HOST): str,
```
vinteo, line 39: MQTT options here but it would probably work better in `OptionsFlow` and can be turned on and off?

custom_components/opensprinkler/__init__.py
Nick-Adams-AU: So would this use the HA core MQTT integration rather than a separate connection to MQTT via the OS integration?

vinteo: No right no it uses MQTT connection based on the MQTT options when you setup the integration, I haven;t figured out how to use the HA core one yet. I just wanted to see if this will work as a proof of concept.

vinteo:
This actually seemed to work pretty well with initial testing, obviously the code still needs a lot of work

travisghansen:
https://github.com/home-assistant/core/blob/dev/homeassistant/components/mqtt_eventstream/__init__.py#L45

Have a look at that or ozw component as well. You’ll need to also add the dependency in manifest.json so it loads in the correct order.

vinteo:
Updated to use the core MQTT integration

custom_components/opensprinkler/__init__.py
vinteo: uses core MQTT integration now

custom_components/opensprinkler/config_flow.py
```diff
@@ -27,6 +33,7 @@
         vol.Required(CONF_PASSWORD): str,
         vol.Required(CONF_MAC): str,
         vol.Optional(CONF_NAME, default=DEFAULT_NAME): str,
+        vol.Optional(CONF_MQTT, default=DEFAULT_MQTT): bool,
```
vinteo, line 36: should this be in `OptionFlow`?

custom_components/opensprinkler/__init__.py
```diff
@@ -47,12 +51,25 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
 
     url = entry.data.get(CONF_URL)
     password = entry.data.get(CONF_PASSWORD)
+    mqtt_option = entry.data.get(CONF_MQTT, DEFAULT_MQTT)
+
     try:
         controller = OpenSprinkler(url, password)
         controller.refresh_on_update = False
 
+        if mqtt_option:
```
travisghansen, line 60: Based on the little I looked at from the other components we’ll probably need more stringent checks here to make sure the install actually has mqtt enabled.

custom_components/opensprinkler/__init__.py
```diff
@@ -47,12 +51,25 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
...
+
+            async def on_message(message):
+                await hass.async_add_executor_job(controller.refresh)
+
+            await mqtt.async_subscribe(hass, MQTT_TOPIC, on_message)
+
         async def async_update_data():
             """Fetch data from OpenSprinkler."""
+
+            if mqtt_option:  # skip refresh
```
travisghansen, line 70: Since mqtt attributes are a subset of the data we can’t shit this down completely can we? Maybe we can up the interval quite high but there are still things that aren’t sending mqtt updates that could hypothetically get out of sync I think.

custom_components/opensprinkler/__init__.py
```diff
@@ -109,9 +126,10 @@ async def async_unload_entry(hass: HomeAssistant, entry: ConfigEntry):
 class OpenSprinklerEntity(RestoreEntity):
     """Define a generic OpenSprinkler entity."""
 
-    def __init__(self, entry, name, coordinator):
+    def __init__(self, entry, name, coordinator, controller):
```
travisghansen, line 129: Maybe we should remove this signature altogether and pull the controller and coordinator from the hass data like line 144 does?

custom_components/opensprinkler/__init__.py
```diff
@@ -150,7 +168,8 @@ async def async_added_to_hass(self):
 
     async def async_update(self):
         """Update latest state."""
-        await self._coordinator.async_request_refresh()
+        await self.hass.async_add_executor_job(self._controller.refresh)
+        self.async_write_ha_state()
```
travisghansen, line 172: If we don’t use the coordinator I'm not entirely sure how the entities get notified that updated values are present. Does invoking this accomplish that?

custom_components/opensprinkler/config_flow.py
travisghansen: No idea honestly :(

custom_components/opensprinkler/__init__.py
vinteo: I think the main thing that would go out of sync is the enable/disable for controller/programs/stations. What should the interval be though? 1 min? 5 min? longer?

custom_components/opensprinkler/__init__.py
vinteo: Yes

vinteo:
For multiple instances, a MQTT update from one device will refresh all instances so technically it wouldn't "break" anything. But yeah, not sure why the topics were namespaced that way.

custom_components/opensprinkler/__init__.py
vinteo: I thought about this a little and I feel like this is an "advanced" user option, so really its up to the user to make sure their setup is correct. Even if we check that the MQTT integration is enabled and check the OS config that MQTT is setup/enabled, there is no guarantee that it would work (even if we compare the two). Of course we should document this somewhere.

Which is why I am thinking it should be an option instead, which the user can easily enable/disable and switch between MQTT and the coordinator. We could even check for the firmware version as well as the OS MQTT config is enabled, otherwise disable/hide the option.

custom_components/opensprinkler/__init__.py
vinteo: set to 5 mins for now

custom_components/opensprinkler/binary_sensor.py
```diff
@@ -94,19 +84,20 @@ def unique_id(self) -> str:
 
     def _get_state(self) -> int:
         """Retrieve latest state."""
-        return bool(getattr(self._controller, self._attr))
+        controller = self.hass.data[DOMAIN][self._entry.entry_id]["controller"]
```
travisghansen, line 87: Should this use the newly created method?

custom_components/opensprinkler/__init__.py
travisghansen: Ok, I just noticed I think they were doing some pretty simple checks in those other integrations. If not yeah it’s advanced for sure so can probably leave as is.

vinteo:
Only thing left is I wanna try get it into options

travisghansen:
I’ve been thinking about this implementation a little and I’m not sure but I think the coordinator will throttle things down as well. We may want the mqtt handler to use the coordinator so if a bunch of mqtt messages come through we don’t slam things similar to how the old integration did for each device.

In essence have some mechanism to ‘debounce’ a bunch of mqtt messages.

PR #112, opened 2020-06-21
refactor: audit entity attributes

vinteo:
Lots of attributes (especially the ones that change frequently) can cause the HA db to explode if history and recorder is turned on depending on the retention settings.

I went through and removed attributes that we already have entities for and the ones I thought would be least useful. We can probably remove more. 

Happy to hear thoughts on which ones should be kept/removed.


custom_components/opensprinkler/__init__.py
```diff
@@ -232,39 +211,12 @@ def device_state_attributes(self):
...
-        ]:
-            iso_attr = attr + "_iso"
-            timestamp = getattr(controller, attr)
-            if not timestamp:
-                attributes[attr] = None
-                attributes[iso_attr] = None
-            else:
-                attributes[iso_attr] = utc_from_timestamp(timestamp).isoformat()
-
         # station counts
```
vinteo, line 220: I feel like we should remove all these counts as well

sdwilsh:
Should it be recommended that the [influxdb integration](https://www.home-assistant.io/integrations/influxdb/) be enabled with this?

custom_components/opensprinkler/__init__.py
```diff
@@ -177,94 +177,38 @@ def device_state_attributes(self):
         controller = self._controller
         attributes = {"opensprinkler_type": "controller"}
         for attr in [
-            "enabled",
             "firmware_version",
             "hardware_version",
             "hardware_type",
-            "hardware_type_name",
```
vinteo, line 184: I removed a lot of the `*_name` ones as I felt that for automations it better to use the non "friendly" types. Thoughts?

travisghansen:
I’m not a fan of this but I’m not entirely sure how bad it’s messing everyone up. Doesn’t seem to be an issue for me but I’m running mysql with 2 weeks I think...not sure how sqlite handles all this data.

vinteo:
how big is you DB with 2 weeks of data?

vinteo:
Did a quick test, starting from a fresh config with no DB, ran the integration for 5 mins. I did not run any stations so the everything was idle for the whole duration. This is how the size of the sqlite database file (home-assistant_v2.db) ended up.

|        | Initial Size | Final Size | Difference |
|--------|--------------|------------|------------|
| Master | 120KB        | 1,880KB    | 1,760KB    |
| Branch | 120KB        | 184KB      | 64KB       |

The branch only had the initial state, it never inserted any more data after first 10 seconds. A difference of ~1.7MB per 5 mins is 20.4MB per hour or 489.6MB per day.

In terms of the initial state of master, the size of the database was 240KB, so the difference of data is 120KB compared to 64KB, so the reduction of removal of the attributes is almost 50% reduction of a entity state change, averaged out. The total difference will of course depend on how often the state changes, but 50% per state change is still a good optimisation IMO.

travisghansen:
What were the counts?

vinteo:
I knew I forgot something. But I have 46 entities so I assume the count for the branch would be 46. If 46 rows provide 120KB for master then the count for master can be estimated to be  1760 / (120 / 46) = 677

travisghansen:
Ok so logging approximately 7% of then entries resulted in 90% reduction in size (the resulting DB was 10% of master). So in a simple comparison (not perfect for sure, but close enough) of the 90% reduction 93% was due to simply logging less entries.  Why bother removing datapoints that aren’t constantly changing (not what drives the incessant entries) since clearly the majority of the delta is from simply logging less entries (vs attempting to make entries smaller)?

Or said differently, by simply removing the constantly changing datapoints we already reduce things by roughly 93%...why bother removing potentially helpful attributes to squeeze a relatively small reduction out.

Or said differently again, 93% of 90 is 83.7. So just over 6% of the total storage difference came from just raw storage removal. 6% of 90 is 5.4. Hypothetically you’d still get a 84.6% reduction (vs 90%) by simply doing what’s necessary to keep the count down.

Disclaimer: I’m not a stats guy, most of the above is just smoke and probably bad math, but I think the general principle should be clear :)

travisghansen:
It’s just not a very compelling argument to me to say: Let’s remove 50% of the attributes to reduce storage by 90%

VS: Let’s remove 1% of the attributes (whatever the really chatty ones are) and reduce storage by 84.6%

Some more testing may prove my numbers all wrong anyhow though :)

vinteo:
I don't disagree with your numbers and I have never disagreed with the fact that we can reduce the storage significantly without removing most of these attributes. My argument has always been that these attributes provided no value to the user. So removing something that provided no value to the user to give us 50% reduction per entry (even though small in terms of actual numbers) is still a very good optimisation.

vinteo:
Like I said before, I have a very pragmatic view of this. To me its weighing the pros and cons.

Pros of removing these extra non-chatty attributes:
- small reduction in storage
- reduce points of failure, although small and easy, any update to the property (name etc) in the library will need to be updated. Unlikely but there could be cases of needing to handle backwards compatibility for older firmware. So complexity to provide no value (IMO)
- provides no value so its just noise (IMO)

Cons:
- a few users may want it (which I am yet to be convinced of, and we can always include it as requested on a case by case basis)

vinteo:
I hope there is no misunderstanding but just in case, I just wanted to add that I didn't do the initial test to prove any point one way or the other. I wanted to test if this PR did what we wanted it to do, which was reduce storage (and number of db entries) and put some numbers to it. I only added the 50% comment because I knew it was a contention point.

travisghansen:
I’m a pretty casual user and several of those removed I like to be able to see (last reboot, and weather call). The _active items are more important than the counterparts that were left so we’d want to put those as their own binary sensors. Flow rate and water level definitely need to be *somewhere* as well (maybe we already have sensors for all those, don’t recall). I think people will want to know historically when programs are/were running so that could be moved to it’s own sensor as well.

I’m just not convinced that:

a - the amount of storage reduction is worth the time
b - there’s no utility to the items being removed (and I’m pretty casual user)

Having said that, I’m happy to move along with the decision if we merge...we can deal with requests if/when they come.

travisghansen:
I would propose the following alternative:

- identify what we consider rapidly changing attrs (I don’t think I would consider is running one for example but rssi and current draw absolutely) and pull those from the list
- of those that get pulled, create specific entities
- make the attributes only be connected to the `enabled` entities (only a single entity per ‘thing’ which would contain the attrs vs all entities)

I would like to see what the real world reduction is with that...perhaps not enough don’t know. You’d still have a relatively high count of entries in the DB but storage requirements may be quite reduced if not logging all the attrs for the high count entities.

vinteo:
All the ones you mentioned, flow rate, water level and *_active, don't we already have sensors for those? That's why I removed them. Those are probably the most important which is why they have their own sensors but they are redundant here. 

Not sure what you mean by worth the time, its a matter of merging this PR. And is it worth the time to maintain them? We have so limited testing already, like I said its the time that we have to spent to maintain the "complexity" is one of the reasons I want to remove them. Having the additional sensors will be even more time to spend to maintain them. I would rather keep them as attributes, but we should have a reason to keep them. Its not just the storage even though that's the main reason for this PR.

Like I said in the original post I am happy to hear which ones should be kept but there should be a reason, please comment on each individual one so I can put them back.

travisghansen:
I’m not sure if we have sensors for those..was thinking not.

We should shout out in slack and let the mob rule :) I’m a pretty light use-case.

vinteo:
When I asked if we already have them I meant we already have them. I didn't
just remove each one for the sake of it. I put  thought into each one. But
I was hoping some would spot any obvious ones with reason to keep. So far
there has been none.

travisghansen:
I guess let’s make a comment with the comprehensive list of datapoints that would no longer be available in any fashion and we’ll tag a few people asking for comments?

PeteBa:
I would go for a minimum attributes set as HA PR reviews are notorious on "kitchen sink" approaches. Again, it is easier to open up the list based on use-cases rather than trying to argue future proofing. I would also convert any "countdown" attributes to "end time" equivalents if that avoids spamming every second.

Initial thoughts - so dont flame me - of other candidates for removal:
- master_station_ - given there aren't any setters for these at the moment
- sensor_1/2_delayed_ - not sure how these be used by an automation
- all of the program attributes (i.e. odd/even, type etc) - since there are no setters to make this configurable
- basically any configuration option as: 1) without setters you dont need them to initialise an HA configuration screen; 2) without generic automation routines then the end-user already knows the config of their opensprinkler and can factor into their automations themselves

Things I would add back in:
- flow_rate - this is a prerequisite for detecting valve failure or leaking pipework
- last_weather_ - a trigger for automation if the weather service is down and your watering level is not updating
- last_reboot_ - a trigger for notification if your "set and forget" device is unexpectedly rebooting or do you have another mechanism for "unavailable"
- water_level - fairly slow moving, i.e. every two hours, and arguably the key data point for a smart sprinkler

vinteo:
@PeteBa flow_rate and water_level have their own sensors. I can put back the other ones. In principle I would agree with the others, but there may be some resistance.

travisghansen:
If HA wants stuff trimmed down for inclusion I back this 100% to attain upstream status.

For this PR should we only include the attrs on the enabled entities and remove the duplication in that front as well?

vinteo:
I have removed more attributes and put the controller attributes on the switch only. Please review again

custom_components/opensprinkler/switch.py
```diff
@@ -84,6 +85,86 @@ def icon(self) -> str:
...
+        ]:
+            iso_attr = attr + "_iso"
+            timestamp = getattr(controller, attr)
+            if not timestamp:
+                attributes[attr] = None
+                attributes[iso_attr] = None
+            else:
+                attributes[iso_attr] = utc_from_timestamp(timestamp).isoformat()
+
+        # station counts
```
vinteo, line 131: Just to ask again, should we remove these station/program counts?

custom_components/opensprinkler/switch.py
```diff
@@ -84,6 +85,86 @@ def icon(self) -> str:
...
+            "last_successfull_weather_call",
+            "last_reboot_time",
+        ]:
+            iso_attr = attr + "_iso"
+            timestamp = getattr(controller, attr)
+            if not timestamp:
+                attributes[attr] = None
+                attributes[iso_attr] = None
+            else:
+                attributes[iso_attr] = utc_from_timestamp(timestamp).isoformat()
```
vinteo, line 129: Do we need both time formats? can we settle with one?

custom_components/opensprinkler/switch.py
travisghansen: Probably can be done with templates...remove.

custom_components/opensprinkler/switch.py
travisghansen: Probably just iso I’d assume as the best option for integrating with hass?

vinteo:
I think we should only do that if we needed to put it in more than one entity, its only one and I don't think it will change

custom_components/opensprinkler/__init__.py
```diff
@@ -412,36 +273,25 @@ def device_state_attributes(self):
...
             "running_program_id",
             "seconds_remaining",
-            "start_time",
-            "max_name_length",
-            "master_1_operation_enabled",
-            "master_2_operation_enabled",
             "rain_delay_ignored",
             "sensor_1_ignored",
             "sensor_2_ignored",
             "sequential_operation",
```
PeteBa, line 282: Curious how `sequential_operation` attribute would be used ?

custom_components/opensprinkler/__init__.py
```diff
@@ -176,136 +176,6 @@ def state(self):
...
-            "last_reboot_cause_name",
-            "water_level",
-        ]:
-            try:
-                attributes[attr] = getattr(controller, attr)
-            except:
-                pass
-
-        for attr in [
-            "last_run_end_time",
```
PeteBa, line 258: `last_run_end_time` is a useful diagnostic value as I often check this to see if the schedule is working correctly. A useful automation would be to notify if there hasnt been any runs in say the last week.

custom_components/opensprinkler/switch.py
```diff
@@ -84,6 +85,46 @@ def icon(self) -> str:
...
+        controller = self._controller
+        attributes = {"opensprinkler_type": "controller"}
+        for attr in [
+            "firmware_version",
+            "hardware_version",
+            "hardware_type",
+            "last_run_station",
+            "last_run_program",
+            "last_run_duration",
+            "rain_sensor_active",
```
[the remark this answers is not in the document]
vinteo: we do "support" older firmware

custom_components/opensprinkler/__init__.py
```diff
@@ -412,36 +273,25 @@ def device_state_attributes(self):
         for attr in [
             "name",
             "index",
-            "enabled",
-            "is_running",
             "is_master",
             "running_program_id",
             "seconds_remaining",
```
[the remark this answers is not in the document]
vinteo: it would not be every second but every 5 seconds, we can change it in the library so maybe not in this PR

custom_components/opensprinkler/__init__.py
vinteo: It has its own sensor

custom_components/opensprinkler/__init__.py
vinteo: Don't know happy to remove

custom_components/opensprinkler/switch.py
```diff
@@ -84,6 +85,46 @@ def icon(self) -> str:
...
+            "sensor_1_enabled",
+            "sensor_2_enabled",
+            "rain_sensor_enabled",
+            "flow_sensor_enabled",
+            "soil_sensor_enabled",
```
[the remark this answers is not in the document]
vinteo: happy to remove

custom_components/opensprinkler/__init__.py
vinteo: removed all the _ignored ones as well

custom_components/opensprinkler/__init__.py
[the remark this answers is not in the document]
vinteo: https://github.com/vinteo/py-opensprinkler/issues/43

custom_components/opensprinkler/switch.py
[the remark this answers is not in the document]
PeteBa: `sensor_1` supports fallback to `rs` in pyopensprinkler [here](https://github.com/vinteo/py-opensprinkler/blob/fecd89ec4b8ed3a91afdc39e066c07c403da29db/pyopensprinkler/__init__.py#L623)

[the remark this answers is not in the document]
vinteo: great, will remove

custom_components/opensprinkler/__init__.py
PeteBa: Resolved

custom_components/opensprinkler/__init__.py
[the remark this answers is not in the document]
PeteBa: Resolved in lieu of 43.

vinteo:
I'll merge this tomorrow morning AU time if there are no more comments

travisghansen:
Sounds good. I’ll merge it all locally and have a look.

vinteo:
Merging...

PR #113, opened 2020-06-21
feat: make mac address optional for new firmware

vinteo:


custom_components/opensprinkler/__init__.py
```diff
@@ -55,7 +55,11 @@ async def async_update_data():
             """Fetch data from OpenSprinkler."""
             _LOGGER.debug("refreshing data")
             async with async_timeout.timeout(TIMEOUT):
-                await hass.async_add_executor_job(controller.refresh)
+                try:
```
vinteo, line 58: Added here to try catch any connection/update exceptions

Issue #120, opened 2020-07-14
last_reboot_time spamming logs


Gibby:
On 0.112.4 and 1.0.7 of the plugin, I see this in my log file about every 5 seconds.

`2020-07-14 00:56:15 ERROR (MainThread) [homeassistant.core] Error doing job: Task exception was never retrieved
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 111, in _handle_refresh_interval
    await self.async_refresh()
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 177, in async_refresh
    update_callback()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 290, in async_write_ha_state
    self._async_write_ha_state()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 317, in _async_write_ha_state
    attr.update(self.device_state_attributes or {})
  File "/config/custom_components/opensprinkler/switch.py", line 104, in device_state_attributes
    timestamp = getattr(controller, attr)
  File "/usr/local/lib/python3.7/site-packages/pyopensprinkler/__init__.py", line 900, in last_reboot_time
    return self._timestamp_to_utc(self._get_variable("lupt"))
  File "/usr/local/lib/python3.7/site-packages/pyopensprinkler/__init__.py", line 343, in _timestamp_to_utc
    return timestamp if timestamp == 0 else timestamp - offset
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'
`

prndlrm:
I have the same problem. My log was growing to 20MB overnight. I uninstalled the integration for now to get it to stop.

vinteo:
Looks like firmware 2.1.7 and below does not have this property. Fix incoming: https://github.com/vinteo/py-opensprinkler/pull/49

vinteo:
Please try upgrading to 1.0.8

Gibby:
@vinteo thanks, working for me

PR #121, opened 2020-07-15
chore: update pyopensprinkler to 0.6.10

vinteo:
Fixes: #120

### DIFF
--- a/custom_components/opensprinkler/manifest.json
+++ b/custom_components/opensprinkler/manifest.json
@@ -2,7 +2,7 @@
   "domain": "opensprinkler",
   "name": "OpenSprinkler",
   "documentation": "https://github.com/vinteo/hass-opensprinkler",
-  "requirements": ["pyopensprinkler==0.6.9"],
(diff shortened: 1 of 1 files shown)


custom_components/opensprinkler/binary_sensor.py
```diff
@@ -136,9 +136,9 @@ def unique_id(self) -> str:
     def icon(self) -> str:
         """Return icon."""
         if self._program.is_running:
-            return "mdi:timer"
+            return "mdi:timer-outline"
```
vinteo, line 139: while testing noticed the deprecation notice in the logs

`2020-07-15 17:49:05 WARNING (MainThread) [frontend.js.latest.202007140] Icon mdi:timer was renamed to mdi:timer-outline, please change your config, it will be removed in version 0.115`

vinteo: actually i am not sure this has been released yet (as i tested it on latest commit in the `core` repo), so I will remove it for now

PR #132, opened 2020-08-17
fix: correct flow rate unit

vinteo:


custom_components/opensprinkler/sensor.py
```diff
@@ -147,7 +147,7 @@ def unique_id(self) -> str:
     @property
     def unit_of_measurement(self):
         """Return the unit of the flow rate."""
-        return "liters/pulse"
+        return "L/min"
```
vinteo, line 150: should this be `L` or `l`?

vinteo: how about `ℓ` instead?

PeteBa: Others use L for liters so would agree with "L/min".

PR #150, opened 2021-06-13
feat: expose firmware version as device attribute

davidjb:
This PR exposes the controller's firmware version as part of the device attributes of the controller operation switch; useful for inclusion within templates, automations and UI.


vinteo:
LGTM, but should we add `firmware_minor_version` too?

davidjb:
Thanks for the suggestion @vinteo - I've added this too and pushed a new commit.

vinteo (approved):
Thank you

PR #188, opened 2022-10-20
Add configuration url

pail23:
Adding the configuration url allows to access the open sprinkler web frontend directly from homeassistant.

### DIFF
--- a/custom_components/opensprinkler/__init__.py
+++ b/custom_components/opensprinkler/__init__.py
@@ -187,6 +187,7 @@ class OpenSprinklerEntity(RestoreEntity):
             "identifiers": {(DOMAIN, slugify(self._entry.unique_id))},
             "name": self._name,
             "manufacturer": "OpenSprinkler",
+            "configuration_url":  f"http://{controller.ip_address}",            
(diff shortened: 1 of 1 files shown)


custom_components/opensprinkler/__init__.py
```diff
@@ -187,6 +187,7 @@ def device_info(self):
             "identifiers": {(DOMAIN, slugify(self._entry.unique_id))},
             "name": self._name,
             "manufacturer": "OpenSprinkler",
+            "configuration_url":  f"http://{controller.ip_address}",
```
vinteo, line 190: Instead of using the controller details here from the API, I think we should be using the URL from the HA configuration, `self._entry.data.get(CONF_URL)` works here.

This means it will have the protocol included and it will work for those that use a proxy or similar

vinteo (changes requested):

vinteo:
I added this change with you as a co-author here: https://github.com/vinteo/hass-opensprinkler/pull/191

It has been released as part of v1.1.11

Thanks for your contribution

PR #230, opened 2023-07-22
feat: Station durations

EdLeckert:
New feature to support viewing and editing program station durations using the Number entity.

I followed your pattern for creating entity types, and created an OpenSprinklerNumber class in __init__.py; but as the Number Entity documentation specified to use "native_value" for the number's value, I didn't see the need for creating a "state" property here, so I left it empty, and just return native_value in number.py. Perhaps you could explain why you decided to use this extra class to return the state in other entities, and let me know how you want to handle this going forward. It will also be an issue in future PRs.

Also, I chose to support minutes instead of seconds. While providing seconds gives the user more control, I don't see any real value in running a zone for under a minute. I do, however, see seconds as being awkward on the front end. Imagine a number box with up and down arrows showing "1500": typing a value would be the only reasonable choice. And if I'm looking at "1500", the first thing I'm going to do is convert it to minutes, which is much easier to comprehend in the context of a zone duration than a large number of seconds. Anyway, let me know how you feel about this.

Requires py-opensprinkler 0.7.7, obviously.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 __pycache__
+.vs/
--- a/custom_components/opensprinkler/__init__.py
+++ b/custom_components/opensprinkler/__init__.py
@@ -33,7 +33,7 @@ from .const import (
 
 _LOGGER = logging.getLogger(__name__)
 
-PLATFORMS = ["binary_sensor", "sensor", "switch"]
+PLATFORMS = ["binary_sensor", "number", "sensor", "switch"]
 TIMEOUT = 10
 
 
@@ -222,6 +222,10 @@ class OpenSprinklerSensor(OpenSprinklerEntity):
     def state(self):
         """Return the state of the sensor."""
         return self._get_state()
+
+
+class OpenSprinklerNumber(OpenSprinklerEntity):
+    """Define a generic OpenSprinkler number."""
 
 
 class OpenSprinklerControllerEntity:
--- a/custom_components/opensprinkler/number.py
+++ b/custom_components/opensprinkler/number.py
@@ -1,0 +1,91 @@
+"""Component providing support for OpenSprinkler number entities."""
+import logging
+from typing import Callable
+
+from homeassistant.components.number import NumberEntity
+from homeassistant.const import CONF_NAME
+from homeassistant.core import HomeAssistant
+from homeassistant.util import slugify
+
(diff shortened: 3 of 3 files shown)


vinteo:
Honestly, in regards to the state pattern I probably copied it from another integration, I can't remember why I did it. I haven't looked at HA dev in a whie so I am not sure what the best practise is now. But what you have looks fine.

Also the `number` entity definitely didn't exist when I wrote this integration, it looks like a nice addition. Would it be worth it to update the switch example in the README to use this `number` entity instead of having to create the `input_number`?

vinteo (approved):

EdLeckert:
If you look at the code for the base entities (e.g. binary_sensor) you'll notice that the "state" property is marked with "@final" decorator , indicating that it should not be overridden. I suppose it's possible that this will be enforced in the future, but it isn't now. The preferred property to implement varies depending on the entity type. For example, it's "is_on" for the binary_sensor.

Most of the code I've just written for hass-opensprinkler uses newer types, some very new. They're very handy for editable fields.

The README will need some work for the new features, but I do have more to push your way if you're up for it.

vinteo:
Sounds good, if you think the old entities should be refactored, go ahead
and submit a PR as well.

For the README I think would be nice to add to it as we go so we don't miss
too much. So the updated switch example for this change would be nice

EdLeckert:
Agreed on the documenting as we go, but I'm not sure that what I implemented is applicable to the example. My changes are for station durations by program, and the example is for an on-demand run of a single station. If there is more than one program, which program would the template use? Even if there's only one program, I may want to use a different duration from the standard program. I wouldn't want to change the program's duration and risk forgetting to reset it. It's more work to use than what you already have in that case. And, you'd still need one template per station...all you'd lose is the input_numbers.

What I would propose is that we implement a new feature for the run-once program, much like in the OpenSprinkler UI. We provide a Number entity for each station and use the **Start Run-Once Program [Keyword /cr]** API feature to implement it. I'd be happy to do that once my other proposed changes are disposed of. And if you want me to document the durations feature now, I'd be happy to do that.

Am I missing something?

vinteo:
Yea sorry I thought what you are doing here is what you are proposing.

I think the README is fine for now then, it is pretty barebones right now
anyway.

PR #232, opened 2023-07-23
Feature: Select odd-even restriction

EdLeckert:
New feature to allow selection of odd-even restriction. Uses Select entity type.


EdLeckert:
Done with this one.

vinteo (approved):

PR #235, opened 2023-07-24
feat: Edit Interval, Starting In Days

EdLeckert:
New feature to view/edit **Interval Days** and **Starting In** Days.

I struggled a bit with the "mode" property. I originally had it set to "box" because that's what I prefer, but the docs suggest setting it to "auto", which seems to give the same result as "slider", the default. The HA UI has little to no support for adjusting this that I can find. However, I discovered that the Mushroom Number card has support for setting this to "Slider" or "Buttons" (similar to "box" but without ability to type a value). So I pulled these values out of the Entities card I'm using on my panel and tried the Mushroom approach, which looks like it will work for me. 

Note that this will affect the Station Durations as well. The are set to "auto" currently.

Thoughts? Do you know of any other cards that allow choosing a different mode? What should our default be?

### DIFF
--- a/custom_components/opensprinkler/number.py
+++ b/custom_components/opensprinkler/number.py
@@ -36,6 +36,10 @@ def _create_entities(hass: HomeAssistant, entry: dict):
                 ProgramDurationNumber(entry, name, program, station, coordinator)
             )
 
+    for _, program in controller.programs.items():
+        entities.append(ProgramIntervalDaysNumber(entry, name, program, coordinator))
+        entities.append(ProgramStartingInDaysNumber(entry, name, program, coordinator))
+
     return entities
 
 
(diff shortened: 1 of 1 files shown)


EdLeckert:
I also have to add some business logic and set min/max, so this is not ready to go quite yet.

vinteo:
I think i prefer it to go with the recommended `auto`. I think it might be possible to have a integration config that the user can set (like for HACS screenshot below) and we can use that value but I think that will require a lot more work.

image

For now like you said the Mushroom custom card could provide an alternative for now.

EdLeckert:
OK, sounds good on the mode.

I pushed another version with min and max set dynamically. The logic mimics what the web UI does to prevent **Starting In** from going above **Interval Days** minus one. Let me know if you know of a better way to check the other value's setting.

Once you're happy I'll do a final squash.

vinteo (approved):
Verified the min/max logic with the web ui

EdLeckert:
Is there something else I need to do before you can merge this?

Issue #237, opened 2023-07-26
Fix hassfest errors


vinteo:
Currently the hassfest build action is throwing the following errors (and warning):

```
Error: R] [SERVICES] Service run has no name and is not in the translations file
Error: R] [SERVICES] Service stop has no name and is not in the translations file
Error: R] [SERVICES] Service set_water_level has no name and is not in the translations file
Error: R] [SERVICES] Service reboot has no name and is not in the translations file
Warning: G] [CONFIG_SCHEMA] Integrations which implement 'async_setup' or 'setup' must define either 'CONFIG_SCHEMA', 'PLATFORM_SCHEMA' or 'PLATFORM_SCHEMA_BASE'. If the integration has no configuration parameters, can only be set up from platforms or can only be set up from config entries, one of the helpers cv.empty_config_schema, cv.platform_only_config_schema or cv.config_entry_only_config_schema can be used.
```

Resolve by updating the service and translation configs.

vinteo:
Fixed in #243 and #239

PR #239, opened 2023-08-07
Move name and description to translations

EdLeckert:
Move service name and description strings to translations folder to support new "Translated services" feature in 2023.8.

Not sure that it's compatible with earlier versions of HA, but I believe the settings will stick until overridden, so probably only an issue for new installations of the integration.

Required change to satisfy "hassfest" Action.

### DIFF
--- a/custom_components/opensprinkler/services.yaml
+++ b/custom_components/opensprinkler/services.yaml
@@ -1,26 +1,18 @@
 run:
-  description: Runs a controller program or station
   fields:
     entity_id:
-      description: Sensor or switch entity id for programs or stations
       example: "sensor.station_name"
     run_seconds:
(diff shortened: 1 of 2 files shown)


vinteo:
Sorry, on holiday so this is the first time I turned on my laptop in more than a week lol. 

Just wanted to confirm that the string have to be removed from the YAML and it cannot be in both places?

EdLeckert:
Well, your timing is excellent, as I just posted this 2 hours ago.

I thought about leaving them for the case where someone on a back version just adding the integration would need them, but thought the case would be rare. Obviously if we did that it would have to be removed eventually. I doubt there's any reason they couldn't be in both, though.

vinteo:
Fixes https://github.com/vinteo/hass-opensprinkler/issues/237

vinteo (approved):

PR #242, opened 2023-08-22
chore-cleanup-translations

EdLeckert:
Cleanup of translations and minor adjustments. Looks worse than it is because strings.json was overwritten (and can be ignored).

1. While I couldn't find a definitive reference in the "official" docs, the consensus seems to be that strings.json should be exactly the same as the primary language in the translations folder, en.json. This is probably for backward compatibility, and the duplicate hopefully can eventually go away. Have copied en.json to strings.json.
2. Error strings do not get translated in non-core integrations. Converted them to the text in core: homeassistant/strings.json
3. Slightly out of scope, but errors for badly formatted URLs during config flow were falling thru to the generic Exception and getting displayed as "Unexpected exception". Now trap the aiohttp InvalidURL exception and display message showing example formatting.

Here's the existing behavior for a "cannot_connect" error, as in wrong IP address:

NotTranslated

New behavior:

FailedToConnect

Also for invalid url formatting:

URLIsMalformed

Finally, invalid authentication:

InvalidAuthentication

Let me know what you think.

### DIFF
--- a/custom_components/opensprinkler/config_flow.py
+++ b/custom_components/opensprinkler/config_flow.py
@@ -2,6 +2,7 @@
 import logging
 
 import voluptuous as vol
+from aiohttp.client_exceptions import InvalidURL
 from homeassistant import config_entries
 from homeassistant.const import CONF_MAC, CONF_NAME, CONF_PASSWORD, CONF_URL
 from homeassistant.helpers.aiohttp_client import async_get_clientsession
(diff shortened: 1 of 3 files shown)


vinteo (approved):
I was hoping they would "fix" the error code translation in non core but its been years and they haven't done it so I guess we just have to copy it which is a shame.

Issue #238, opened 2023-08-03
Issue upgrading to Home Assistant Core 2023.8.0


deanjunk:
Logger: homeassistant.setup
Source: setup.py:209
First occurred: 12:03:32 PM (1 occurrences)
Last logged: 12:03:32 PM

Setup failed for custom integration opensprinkler: Requirements for opensprinkler not found: ['pyopensprinkler==0.7.6'].

vinteo:
Are you on version 1.1.17?

deanjunk:
No, v1.1.15 which is the latest version available through HACS.

vinteo:
It should be v1.1.17. did you update HACS as well? Not sure why yours is
still at 1.1.15

deanjunk:
Everything (except Home Assistant Core) is updated and the latest version is v1.1.15...

vinteo:
The only way to fix your issue is to upgrade so you might need to upgrade
manually. There are instructions in the README

luebbe:
Same here. First after upgrading to HASS 2023.8.0 I was also confronted with the "requirements not found" error message.
One or two HACS versions later, these errors disappeared, but the OpenSprinkler integration still doesn't work. Everything is flagged as unavailable.

Today I updated to HASS 2023.8.4 and the error message upon startup is (repeated multiple times)

> Dieser Fehler wurde von einer benutzerdefinierten Integration verursacht
> 
> Logger: custom_components.opensprinkler
> Source: helpers/update_coordinator.py:262
> Integration: OpenSprinkler (documentation, issues)
> First occurred: 15:23:00 (6 occurrences)
> Last logged: 15:26:32
> 
> Error fetching OpenSprinkler resource status data: Error fetching OpenSprinkler state

OpenSprinkler is installed via HACS and I am on version 1.1.17 since a while.
I can connect to open sprinkler without any problems using the app or a web browser.

How can I get the open sprinkler integration up and running again?

EdLeckert:
Today I installed a clean copy of Home Assistant 2023.8.4, HACS, and OpenSprinkler 1.1.17 (via HACS) on a clean VM. It works fine.

The only way I've been able to re-create your exact error is to go into .storage and corrupt the password. Then I get your error, but I also get:

```
Logger: homeassistant.config_entries
Source: config_entries.py:430
First occurred: 3:34:32 PM (1 occurrences)
Last logged: 3:34:32 PM

[...]

PR #243, opened 2023-08-23
chore-eliminate-hassfest-warning

EdLeckert:
Addresses 2nd part of "Fix hassfest errors #237", the Warning:

[CONFIG_SCHEMA] Integrations which implement 'async_setup' or 'setup' must define either 'CONFIG_SCHEMA', 'PLATFORM_SCHEMA' or 'PLATFORM_SCHEMA_BASE'. If the integration has no configuration parameters, can only be set up from platforms or can only be set up from config entries, one of the helpers cv.empty_config_schema, cv.platform_only_config_schema or cv.config_entry_only_config_schema can be used.

May possibly address "Issue upgrading to Home Assistant Core 2023.8.0 #238". I don't use HACS, so don't know if it's still stuck at v1.1.15, but if it is, this could possibly be the cause, if the Warning above is the result of a new rule. Not sure when the rule was introduced or if Warnings prevent upgrades. Have no idea how that works, actually.

In any case, is it safe to assume that since we use Config Flow that we no longer support yaml configuration and also don't need async_setup?

### DIFF
--- a/custom_components/opensprinkler/__init__.py
+++ b/custom_components/opensprinkler/__init__.py
@@ -39,11 +39,6 @@ PLATFORMS = ["binary_sensor", "number", "select", "sensor", "switch", "text", "t
 TIMEOUT = 10
 
 
-async def async_setup(hass: HomeAssistant, config: dict):
-    """Set up the openSprinkler component from YAML."""
-    return True
-
-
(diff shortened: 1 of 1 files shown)


vinteo (approved):
I don't think had any issue updating in HACS with 2023.08

But yea I think `async_setup` was needed in the transition to config flow in the past but its probably not needed anymore

PR #244, opened 2023-08-26
feat-display-concise-errors

EdLeckert:
If the password or IP address changes or gets corrupted, or any other error occurs while trying to refresh data, the exact cause of the error is not shown. This complicated the investigation of the second complaint by @deanjunk in #238  **Issue upgrading to Home Assistant Core 2023.8.0**.

The current error message for any type of error is:

InvalidPasswordDetail_2

With the proposed changes, for an invalid password the error log shows:

InvalidPasswordSummary

And the detail of the error is:

InvalidPasswordDetail

Finally, on a corrupted or simply changed IP address, the error log is:

InvalidURLDetail

While the proposed changes don't technically "handle" the error, there's really not much else that can be done. And the statement:

```
                if not controller._state:
                    raise UpdateFailed("Error fetching OpenSprinkler state")

```

seems to catch everything anyway.


vinteo (approved):

EdLeckert:
I noticed you approved but did not merge this PR. Is there something I need to do? Do you want me to merge it?

PR #247, opened 2023-09-15
fix-docs-port-default

EdLeckert:
The default port for OpenSprinkler is "80", though the `README.md` shows "8080". This is the case at least as far back as the earliest available User Manual, for Firmware 2.1.6.

From the latest Firmware 2.2.0 manual:

> Next, open a browser and type in the
> URL http://os-ip (e.g. http://192.168.1.122). In case you have set a custom HTTP port number
> (other than the default 80), include the port number in the URL as well.


EdLeckert:
I see from old Issues and posts that the default port was 8080 back in the day. I can change the message to address that if you can give me more specific info about when it changed.

vinteo (approved):

Issue #249, opened 2023-09-20
Integration calls 'OpenSprinkler Current Time changed to xxxx' Infinite Loop every 5 seconds


johnny-co:
Reviewing the HomeAssistant Logbook in device details for my OSPI OpenSprinkler (Firmware 2.2.0 (2)) a call is being made to set current time every 5 seconds indefinitely.

OpenSprinkler Current Time changed to September 20, 2023 at 8:54 AM
8:54:23 AM - 1 second ago
OpenSprinkler Current Time changed to September 20, 2023 at 8:54 AM
8:54:18 AM - 6 seconds ago
OpenSprinkler Current Time changed to September 20, 2023 at 8:54 AM
8:54:13 AM - 11 seconds ago
OpenSprinkler Current Time changed to September 20, 2023 at 8:54 AM
8:54:08 AM - 16 seconds ago

EdLeckert:
Yes, it is a bit chatty, as the integration refreshes every 5 seconds.

While we work on a better solution you can disable the entity. Go to Settings\Devices & Services\OpenSprinkler, click on the device. Under the `Sensors` section click on `Current Time`, click on the gear icon, turn off `Enabled` and click Update. The entity will no longer be tracked.

johnny-co:
Thank you that worked.

vinteo:
Entity is disabled by default in v1.2.1.

Issue #248, opened 2023-09-20
1.2.0 Update rendered implementation unusable


theOrakleAtBurns:
I love the innovation and the additions, but I have 40 station and 10 programs...

I'll let you do the math, but this created more entries than Hass could handle with your current threading.

EdLeckert:
Well, that's an impressive system you have there.

Were you able to roll back to the previous version while we figure this out?

Also, may I ask what features of the integration you mostly use?

EdLeckert:
Sorry you had to deal with that. I've had HA upgrades blow up my system and it's no fun.

May I ask what you're using for your HA server?

Also, can you tell me a bit more about your experience with the update...briefly, what worked and what didn't?

theOrakle:
No prob at all man - so hass runs on a ESX cluster of NUCs  (i7,32 GB) connected to an all flash synology array via 10 G iSCSI.  The db is Postgres on another VM on the same cluster. Each node has 2 core & 8GB.

On the upgrade front:
 - it was the only custom component that was ungraded
 -  Watched the logs after restart & it complained that a value needed to be 0-23
 - navigated to the integration via hass just fine
 - One I clicked on the device, it couldn’t render the 630 entities in under 2 minutes and scrolling was not responsive on Safari, Chrome or companion app

theOrakle:
Note - I have more entities on other integrations with more devices.   


![IMG_3015](https://github.com/vinteo/hass-opensprinkler/assets/37029980/701df286-fe40-4624-ba7c-577b0f4ea2f1)

theOrakle:
Forgot to add… ran top on both hass and the db while I was waiting for the integration to render one of the times and no I/o or cpu wait

theOrakle:
I have another environment (yellow with a gGb/s NvME), that I can use if you want me to test something for you

theOrakle:
You are correct sir.

EdLeckert:
OK, I set up an environment to simulate your system as best as I can with only 8 stations and 1 program in my one controller. I modified the integration to create entities for 10 copies of the one program and 5 copies of the 8 stations, for a total matching your system. This resulted in the creation of 630 entities for the 1 device in the integration. I'm running VirtualBox on a laptop, Chrome on same laptop, wired connection to my OS controller. Far from an ideal comparison, but gives me an idea of the impact.

[...]

PR #250, opened 2023-09-20
fix-disable-current-time-by-default

EdLeckert:
Disables `sensor.opensprinkler_current_time` by default. Addresses Issue #249.

Was planning to remove this entity and replace with a service, but now that it's released I think this is the way to go. Can still add a service later.

We don't need a constant update of the current time, but I have found it to wander in the past unexpectedly, so it's nice to have a way to check periodically.

### DIFF
--- a/custom_components/opensprinkler/sensor.py
+++ b/custom_components/opensprinkler/sensor.py
@@ -345,6 +345,11 @@ class ControllerCurrentTimeSensor(
         """Return a unique, Home Assistant friendly identifier for this entity."""
         return slugify(f"{self._entry.unique_id}_{self._entity_type}_devt")
 
+    @property
+    def entity_registry_enabled_default(self) -> bool:
+        """Set entity disabled by default."""
+        return False
+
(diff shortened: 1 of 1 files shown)


vinteo (approved):

vinteo:
should I do a release for this? I am guessing it doesn't help for those who already upgraded though?

EdLeckert:
I doesn't help those who have already upgraded...they can disable it manually as I explained in Issue #249.

However, for those who have _not_ upgraded, it will save them that trouble, so if it's not too much trouble to release it might be worth it. It's up to you. I've been running with it that way for a month or two and it didn't bother me, but some find it annoying. Two have noticed and reported it so far.

Also know that I'm working on a possible fix for #248. I doubt that many installations are anywhere near as large as this one, but I can't know for sure. My idea is to add an option to Config Flow and Options Flow to _not_ create the Program entities, warning users of the number of entities that would be created. Should have something for you to look at in a day or two.

Issue #252, opened 2023-09-23
Error adding entities for domain time with platform opensprinkler when using sunrise/sunset offsets


aav7fl:
If my water programs use sunrise/sunset offsets like in the screenshot below, the hass-opensprinkler component is unable to create program sensors. This blocks my integration from loading and working. 

![image](https://github.com/vinteo/hass-opensprinkler/assets/3487107/511b0089-7fb2-43a3-988c-4bb683db4701)

with this log error:

```
2023-09-22 23:06:33.006 ERROR (MainThread) [homeassistant.components.time] Error adding entities for domain time with platform opensprinkler
Traceback (most recent call last):
File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 507, in async_add_entities
await asyncio.gather(*tasks)
File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 752, in _async_add_entity
await entity.add_to_platform_finish()
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1002, in add_to_platform_finish
self.async_write_ha_state()
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 730, in async_write_ha_state
self._async_write_ha_state()
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 830, in _async_write_ha_state
state, attr = self._async_generate_attributes()
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 771, in _async_generate_attributes
state = self._stringify_state(available)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 736, in _stringify_state
if (state := self.state) is None:
^^^^^^^^^^
File "/usr/src/homeassistant/homeassistant/components/time/__init__.py", line 94, in state
if self.native_value is None:
^^^^^^^^^^^^^^^^^
File "/config/custom_components/opensprinkler/time.py", line 72, in native_value
the_time = datetime.time(round(minutes / 60), minutes % 60, 0)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: hour must be in 0..23
```

[...]

PR #254, opened 2023-09-24
feat-support-sunrise-sunset-start-time

EdLeckert:
Feature to support sunset and sunrise as program start times. Also addresses "ValueError: hour must be in 0..23" error reported in #248 and #252, since the error was caused by lack of support for this feature.

At this time, only `start0` is supported. Additional start times are not yet supported.

This PR replaces #253, which has been closed.


custom_components/opensprinkler/time.py
```diff
@@ -67,10 +76,24 @@ def icon(self) -> str:
...
+            )
+        elif is_set(start0, START_TIME_SUNRISE_BIT):
+            minutes = self._controller.sunrise + (
+                (start0 & START_TIME_MINUTES_MASK) * sign
+            )
+        else:
+            minutes = start0
+
+        start_time = datetime.time(trunc(minutes / 60), minutes % 60, 0)
+        return start_time
```
vinteo, line 96: Are you planning to keep this logic here or move it to pyopensprinkler eventually? 

I was wondering if `program_start_times` should always just return the time and have properties for each bit/flag and even the offset in pyopensprinkler. What do you think?

EdLeckert: Sorry, I was planning to leave a comment about the bit twiddling and got busy. 

But yes, the bit logic belongs in pyopensprinkler where your other bit logic lives. I was just trying to get the issue for sunrise/sunset folks resolved quickly, and given the issues some have had with updating, wanted to keep the releases of pyopensprinkler to a minimum. I figured I would move it later, possibly when adding support for Additional Start Times. 

As to the exact implementation details for moving the logic, I hadn't thought that far ahead, except that the UI shouldn't change. I do find it useful to have the type of offset, the offset amount, and the actual time (directly settable) available to the user. For example, all three of these entities are editable:

StartTime

vinteo (approved):

PR #262, opened 2023-10-16
feat-support-additional-start-times

EdLeckert:
Feature to support additional start times (start1-start3) and use new `py-opensprinkler` PR#85 to encode/decode bits for start times rather than doing it here.

Depends on `py-opensprinkler` PR https://github.com/vinteo/py-opensprinkler/pull/85.

One small breaking change that probably won't affect anyone:

`select.[program-name]_additional_start_times`

has changed to:

`select.[program-name]_additional_start_time_type`

because the old name didn't make sense. Selects options `Repeating` or `Fixed`, and wasn't useful by itself without visiting the OpenSprinker UI. Still isn't until `Repeating` is supported, which should be soon.

These changes allow creation of a view that resembles the OpenSprinkler UI, such as the following:

OpenSprinkler Dashboard

All fields are editable by clicking to invoke a more-info dialog. These are Mushroom Entity cards, but other card types allow in-place editing.

All entities for `Additional Start Times` are disabled by default to avoid entity bloat.


vinteo (approved):

EdLeckert:
@vinteo: I have completed the changes to support `Repeating` `Additional Start Times`, but I'm afraid I don't know how to create a PR that includes only those changes when the previous PR is still pending--the new PR would show those previous changes as well as the new ones.

Can you explain how I can create a PR with only the changes in the new branch against the last PR, as opposed to against Master? Or would it be OK to merge the pending PR into Master before creating the new PR?

vinteo:
Usually you change the base branch from master to do this, but I am not sure that will work from a forked repo.

Since you have contributer status you could work directly on this repo rather than from a forked repo. So create branch and push directly on a this repo. Clone this repo to your local and copy the changes on this branch into a new barnch (branch-a) and opening a new PR. Then create the new branch (branch-b) off that branch and opening another PR but this time change the base branch from master to branch-a. 

Then in the future you can just work off this repo directly locally.

Not sure if I explanied that well, hope it maskes sense.

vinteo:
I just did a py-opensprinkler release so if we create and merge a PR with the new version we can merge this in as well.

EdLeckert:
I just confirmed from the log that I did base the `feat-support-repeating-start-times` branch off of the `feat-support-additional-start-times` branch rather than `master` in my fork, and as you suspected, that doesn't help.

So, I will begin again from the main repo and create two new PRs from there.

EdLeckert:
Done. Thanks for the info!

PR #264, opened 2023-11-07
feat-support-additional-start-times

EdLeckert:
Feature to support additional start times (start1-start3) and use new `py-opensprinkler` PR#85 to encode/decode bits for start times rather than doing it here.

Depends on `py-opensprinkler` PR https://github.com/vinteo/py-opensprinkler/pull/85.

One small breaking change that probably won't affect anyone:

`select.[program-name]_additional_start_times`

has changed to:

`select.[program-name]_additional_start_time_type`

because the old name didn't make sense. Selects options `Repeating` or `Fixed`, and wasn't useful by itself without visiting the OpenSprinker UI. Still isn't until `Repeating` is supported, which should be soon.

These changes allow creation of a view that resembles the OpenSprinkler UI, such as the following:

OpenSprinkler Dashboard

All fields are editable by clicking to invoke a more-info dialog. These are Mushroom Entity cards, but other card types allow in-place editing.

All entities for `Additional Start Times` are disabled by default to avoid entity bloat.


vinteo (approved):

vinteo:
Looks good, let me know when this is ready to go

EdLeckert:
Well, unfortunately there was a bug in my changes to py-opensprinkler. Sorry about that. But with that fix, it all looks good.

vinteo:
released 0.7.10, update it here and we can merge

EdLeckert:
Updated `manifest.json` to use 0.7.10. Downloaded branch `feat-support-additional-start-times` to my site and it looks good.

Also updated branch `feat-support-repeating-start-times` by mistake. Please ignore.

vinteo (approved):

PR #265, opened 2023-11-07
feat-support-repeating-start-times

EdLeckert:
Feature to complete support for `Additional Start Times`. Requires `py-opensprinkler v0.7.9`.

By using Conditional Cards, users can display the appropriate controls, depending on the `Start Time Type`:

AdditonalStartTimes

Because data in the fields start1-3 is stored differently depending on the `Start Time Type`, these fields are cleared by `py-opensprinkler` whenever the type is changed.

The new entities for repeat count and interval are created `Disabled` to reduce entity bloat for those who do not use `Additional Start Times`.


vinteo (approved):

vinteo:
Is this ready to merge into the other branch?

EdLeckert:
Yes, with the caveat that I have not pulled the code back from your repo and put it on my site yet, but it should be exactly the same

PR #267, opened 2023-11-12
feat-support-weekly-schedule

EdLeckert:
Fully support `Interval` vs. `Weekly` scheduling. Creates a new `switch` entity for each day of the week for each program. Entities are disabled by default to avoid entity bloat for those not using `Weekly` scheduling. Depends on `py-opensprinkler` PR https://github.com/vinteo/py-opensprinkler/pull/88.

This completes support for all features available on the older (pre `Date Range`) version of the `OpenSprinkler` `Edit Programs` web page.

A conditional card can be used to display only appropriate controls, depending on the `Program Type` currently selected, as shown here:

CombinedProgramTypes


vinteo:
Have released 0.7.11, please update it on this PR

vinteo (approved):

Issue #251, opened 2023-09-22
Service to set rain delay


vinteo:
pyopensprinkler has the controller function:

https://github.com/vinteo/py-opensprinkler/blob/master/pyopensprinkler/__init__.py#L372-L380

noctividus:
It would be great if we could get this implemented.

The advantage of using a rain delay is that it is temporary. Currently, I use an automation to disable open sprinkler in the event of forecast rain. When that forecast clears, I use a second automation to re-enable open sprinkler.

This isn't ideal, because if HomeAssistant misses the second event, my OpenSprinkler can remain disabled for some time.

If we had the ability to set a rain delay, then I could set a rain delay for 2 days, opensprinkler would automatically re-enable and I wouldn't have to worry about a second automation.

curlydingo:
Happy to give it a go, just getting a dev environment set up now.

curlydingo:
I have it "working" but ran into a bit of trouble, please see draft pull request for details.

vinteo:
Done in v1.2.6 (#269), thanks @curlydingo

PR #269, opened 2024-01-06
Add rain delay service

curlydingo:
Caveat: It's my first time, please be gentle.

Adds a rain delay service to the integration per #251. Modelled largely off the existing service to set the water level.

Tested in a HA dev container against my local opensprinkler instance and works well, BUT I had to do some extra filtering of the EntityPlatforms in order to be able to call the service at all, which the rest of the code doesn't do. At the moment, in my dev environment none of those other services can be called. The specific change is in the second commit.

At first glance it seems like a change from HA has caused a need for this but I can't seem to find it. Grateful for any help that can be offered, but until we figure this out I'm not sure it should be merged in its current state.

Second caveat: The DE and SK translations are literally just an online translator's output of EN; not sure how you'd like these to be done properly?

### DIFF
--- a/custom_components/opensprinkler/__init__.py
+++ b/custom_components/opensprinkler/__init__.py
@@ -26,10 +26,12 @@ from .const import (
     DOMAIN,
     SCHEMA_SERVICE_REBOOT,
     SCHEMA_SERVICE_RUN,
+    SCHEMA_SERVICE_SET_RAIN_DELAY,
     SCHEMA_SERVICE_SET_WATER_LEVEL,
     SCHEMA_SERVICE_STOP,
     SERVICE_REBOOT,
     SERVICE_RUN,
+    SERVICE_SET_RAIN_DELAY,
     SERVICE_SET_WATER_LEVEL,
     SERVICE_STOP,
 )
@@ -134,6 +136,18 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         service=SERVICE_SET_WATER_LEVEL,
         schema=cv.make_entity_service_schema(SCHEMA_SERVICE_SET_WATER_LEVEL),
         service_func=_async_send_set_water_level_command,
+    )
+
+    async def _async_send_set_rain_delay_command(call: ServiceCall):
+        await hass.helpers.service.entity_service_call(
+            async_get_platforms(hass, DOMAIN), SERVICE_SET_RAIN_DELAY, call
+        )
+
+    hass.services.async_register(
+        domain=DOMAIN,
+        service=SERVICE_SET_RAIN_DELAY,
+        schema=cv.make_entity_service_schema(SCHEMA_SERVICE_SET_RAIN_DELAY),
+        service_func=_async_send_set_rain_delay_command,
     )
 
     async def _async_send_reboot_command(call: ServiceCall):
@@ -314,6 +328,11 @@ class OpenSprinklerControllerEntity:
     async def set_water_level(self, water_level: int):
         """Set water level percentage"""
         await self._controller.set_water_level(water_level)
+        await self._coordinator.async_request_refresh()
+
+    async def set_rain_delay(self, rain_delay: int):
+        """Set rain delay hours"""
+        await self._controller.set_rain_delay(rain_delay)
         await self._coordinator.async_request_refresh()
 
     async def reboot(self):
--- a/custom_components/opensprinkler/__init__.py
+++ b/custom_components/opensprinkler/__init__.py
@@ -139,13 +139,8 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
     )
 
     async def _async_send_set_rain_delay_command(call: ServiceCall):
-        for platform in async_get_platforms(hass, DOMAIN):
-            if platform.domain == "sensor":
-                sensor_platform = platform
-                break
-
         await hass.helpers.service.entity_service_call(
-            sensor_platform.entities, SERVICE_SET_RAIN_DELAY, call
+            async_get_platforms(hass, DOMAIN), SERVICE_SET_RAIN_DELAY, call
         )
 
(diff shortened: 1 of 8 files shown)


curlydingo:
FYI my dev environment is a Dev Container in VSCode based on HA/Core/dev. No other integrations installed.

custom_components/opensprinkler/__init__.py
```diff
@@ -136,6 +138,23 @@ async def _async_send_set_water_level_command(call: ServiceCall):
         service_func=_async_send_set_water_level_command,
     )
 
+    async def _async_send_set_rain_delay_command(call: ServiceCall):
+        for platform in async_get_platforms(hass, DOMAIN):
+            if platform.domain == "sensor":
+                sensor_platform = platform
+                break
```
vinteo, line 145: Is this pattern needed, seems different from all the rest

vinteo:
I think if the translations are missing it should just default to EN? so just omit them for now

custom_components/opensprinkler/strings.json
```diff
@@ -66,6 +66,20 @@
...
     },
+    "set_rain_delay": {
+      "name": "Set rain delay",
+      "description": "Set rain delay of the sprinkler.",
+      "fields": {
+        "entity_id": {
+          "name": "Entity",
+          "description": "The rain delay entity to change."
+        },
+        "water_level": {
```
vinteo, line 77: this should be `rain_delay`

custom_components/opensprinkler/__init__.py
vinteo: I just re-read you description, maybe @EdLeckert might have an idea why

EdLeckert:
`services.yaml`

The standard unit_of_measurement for hours is "h", per `.../homeassistant/const.py`

`# Time units`
...
`TIME_HOURS = "h"`

EdLeckert:
@curlydingo, can you better explain what you're trying to accomplish with the filtering code that Vincent points out above. The `selector` in `services.yaml` limits the user's choice of entities to `sensor` types already. Now, some of those entities will work, and some won't, but they're all sensors in any case.

It appears that the reason the code is throwing an error is that `async_get_platforms` returns a `list[EntityPlatform]`, whereas your filtering logic tries to pass `sensor_platform.entities`, a `dict[str, Entity]`, to the service call. This actually works:

```
        await hass.helpers.service.entity_service_call(
            [sensor_platform], SERVICE_SET_RAIN_DELAY, call
        )
```

However, I'm not convinced we need this filtering.

In my test environment, `Set water level` and `Set rain delay` both work with the change above.

Incidentally, this issue of having to guess which entities will work with services that are really device services is something I've been meaning to look into, so don't feel you have to solve it today.

curlydingo:
HI @EdLeckert, @vinteo, thanks for looking in to this.

TLDR; The code that @EdLeckert gave in his response does *not* work for me - it fails with the stack trace below. So something must be different about our setups. Am I on a too-new version of HA's devcontainer somehow?

Firstly, I should highlight that I didn't really want any extra filtering code in the first place, but without it the service call fails. It seems that  _get_permissible_entity_candidates, which HA calls at some stage during the service call, requires a dict[str, Entities] in order to do its checks. Stack trace here: 

_get_permissible_entity_candidates (\workspaces\core\homeassistant\helpers\service.py:781)
entity_service_call (\workspaces\core\homeassistant\helpers\service.py:828)
_async_send_set_rain_delay_command (\workspaces\core\config\custom_components\opensprinkler\__init__.py:148)
_execute_service (\workspaces\core\homeassistant\core.py:2221)
async_call (\workspaces\core\homeassistant\core.py:2184)
run (\workspaces\core\homeassistant\runner.py:188)
main (\workspaces\core\homeassistant\__main__.py:209)
<module> (hass:8)

At line 781 of service.py there is a call to entities.get, which fails if 'entities' is a list rather than a dict.

I get the same exception for services I haven't changed, including setting the water level, so I'm fairly confused.

The code in the pull request is basically just what I came up with to choose an appropriate dict[str, Entities] from the list that async_get_platforms provides, but I don't think any of us like it! (At least until we can figure out why it's necessary.)

curlydingo:
[Found it, I think.](https://github.com/curlydingo/core/commit/09b65f14b96d53812d44eb02f6fe603f1f2c5ace#diff-a67fd3ef25fbcb4a89fc86ce3ed65ae5bd7faab09c10e1ae0ced9f9b7bd22f25L796)

Given the upstream change, should I add similar filtering to the other service calls? Happy to take advice on how to do that - the way I've written it seems a bit non-pythony (it's not my strongest language).

Alternatively, we implement a solution that only provides a single entity to the service call, since as @EdLeckert mentions it doesn't seem like we really care about the entity in these cases anyway - they're really services for the device.

EdLeckert:
I just upgraded my production system to HA 2024.1.2 and `Set water level` works fine.

curlydingo:
Yes, the commit I linked doesn't seem to be in 2024.1.2. I've set my dev environment to that release (instead of core's dev branch) and it works fine without the extra filtering. I've removed it from this PR since the issue doesn't need to be solved yet.

custom_components/opensprinkler/translations/en.json
```diff
@@ -66,6 +66,20 @@
...
     },
+    "set_rain_delay": {
+      "name": "Set rain delay",
+      "description": "Set rain delay of the sprinkler.",
+      "fields": {
+        "entity_id": {
+          "name": "Entity",
+          "description": "The rain delay entity to change."
+        },
+        "water_level": {
```
vinteo, line 77: `rain_delay` here as well

vinteo (approved):
Happy to merge this as well if @EdLeckert approves it as well

EdLeckert (approved):
Tested against HA 2023.10.3 and 2024.1.2

vinteo:
Thanks @curlydingo

Issue #218, opened 2023-06-15
New pause feature


Hootie81:
Last week there was a new version of opensprinkler firmware released, it includes a pause feature, and also setting start and end dates to programs. it also includes a change so if a station is stopped, the remaining stations can be moved forward instead of having to wait for the additional time to pass for the stopped station.

this pause feature I have been eagerly awaiting since I first jumped in the shower with the sprinklers running! to make use of it I really would like to have an automation that pauses the sprinklers when the bathroom is occupied, any chance of adding a service to the integration?

vinteo:
Definitely something I or some one can add. NO promises on timeframe though

vinteo:
available in v1.2.7

PR #270, opened 2024-01-15
Control pause

curlydingo:
Per #218, hopefully better showers for all now :)

I've taken the liberty of incrementing the minor version number, but I'm not sure if that's appropriate or should happen later somehow. Apologies if I got it wrong.

### DIFF
--- a/custom_components/opensprinkler/sensor.py
+++ b/custom_components/opensprinkler/sensor.py
@@ -8,7 +8,6 @@ from homeassistant.core import HomeAssistant
 from homeassistant.helpers.entity import Entity
 from homeassistant.util import slugify
 from homeassistant.util.dt import utc_from_timestamp
-from homeassistant.util.dt import now
 
 from . import (
     OpenSprinklerControllerEntity,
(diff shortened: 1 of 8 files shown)


curlydingo:
Hmm, the linter doesn't seem to like that the new PauseEndTimeSensor has 13 lines of code that match the RainDelayStopTimeSensor.

I could:
1. Make a new class "EndTimeSensor" and derive PauseEndTimeSensor and RainDelayStopTimeSensor from it (seems like overkill), or
2. Change a comment or something small in the middle of that block (seems like a bit of a kludge).

Or another option you might come up with? Happy either way just let me know.

vinteo:
Maybe a different icon (might make sense anyway) might help, but not sure if it be enough

Also I didn't give a minor verion for the other service and usually don't for dep upgrade so a patch version might be ok

vinteo:
looks like I forgot to update the vsersion there so it should be 1.2.7 for a patch

custom_components/opensprinkler/sensor.py
```diff
@@ -240,6 +241,48 @@ def _get_state(self):
...
+        super().__init__(entry, name, coordinator)
+
+    @property
+    def device_class(self):
+        """Return the device class."""
+        return SensorDeviceClass.TIMESTAMP
+
+    @property
+    def icon(self) -> str:
+        """Return icon."""
```
vinteo, line 261: I wonder if linting requires these comments, some of these seem pretty useless. Could help with the duplicate linting error if removed

vinteo: Another hack could be to just change the order of the properties... lol

vinteo (approved):

curlydingo:
Sorry, only just saw the request for Ed's review! Don't release until he's had a look, I suppose?

vinteo:
@curlydingo I forgot that we should add examples in the README. For the other service as well

curlydingo:
Yep, happy to. Might update the tested versions at the top too, unless you object.

PR #273, opened 2024-01-18
Add new services to README and updated version references.

curlydingo:
Added the new services and updated the README version.

The HACS action badge still says failing, but the action is passing. Doing a quick search, it seems like sometimes github can show this because of an old failed run? There is one failure for the master branch in the history, but it's so old the logfiles aren't kept anymore. Should we just delete that run and see if it fixes the badge?


vinteo (approved):

vinteo:
i deleted the only failure for HACS, let's see if it works

curlydingo:
> i deleted the only failure for HACS, let's see if it works

Looks like it worked!

Issue #278, opened 2024-02-10
Calling Service fails


qvqn:
Not sure if it's my install but getting failure errors since going to 2024.2

My automation fails, so I tried to call the service and I get this error

![image](https://github.com/vinteo/hass-opensprinkler/assets/14333303/ef57b056-a134-45d2-bdf7-66f76af68290)

![image](https://github.com/vinteo/hass-opensprinkler/assets/14333303/a787b1fb-bbd0-47c6-b030-0c2aba84852e)

curlydingo:
Ahh here it is. I know what this is, it came up in a PR I raised a while ago but we didn’t know what release it would hit us in. I should be able to fix in the next day or two.

qvqn:
glad you are aware of it, was confused no one raised it since 2024.2 has been released for a bit.

I use the call service as my ha is internet exposed and not my ospi so I can manually turn on stations via 4G to check leaks/blockages when I'm out of wifi range.

curlydingo:
Yeah I hadn’t applied the 2024.2 update for a while after it was released :)

I think a lot of people are in a similar situation to you with their ospi
not being online so am keen to get the fox out ASAP.

vinteo:
Should be fixed in 1.3.0

qvqn:
thank you very much for the update

PR #279, opened 2024-02-10
Fix broken services in HA `2024.2.1`

DylanGriffith:
In Home Assistant `2024.2.1` this service is always erroring with:

```
Failed to call service opensprinkler/run. 'list' object has no attribute 'get'
```

The stacktrace for this error is below:

```
2024-02-10 12:46:23.684 ERROR (MainThread) [homeassistant.components.websocket_api.http.connection] [<REDACTED>] 'list' object has no attribute 'get'
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/components/websocket_api/commands.py", line 240, in handle_call_service
    response = await hass.services.async_call(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2279, in async_call
    response_data = await coro
                    ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2316, in _execute_service
    return await target(service_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/opensprinkler/__init__.py", line 108, in _async_send_run_command
    await hass.helpers.service.entity_service_call(
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 843, in entity_service_call
    entity_candidates = _get_permissible_entity_candidates(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 796, in _get_permissible_entity_candidates
    and (entity := entities.get(single_entity)) is not None
                   ^^^^^^^^^^^^
AttributeError: 'list' object has no attribute 'get'
```

After some investigation I found that the function signature for `entity_service_call` was changed in
https://github.com/home-assistant/core/pull/106759 from `Iterable[EntityPlatform]` to `dict[str, Entity]`.

It's not clear to me if this is some internal API that cannot be relied upon by custom components as it does not seem this was a widely communicated API change. But in any case it seems this change fixes the problem for me.

This new function `async_get_entities` was copied from a similar change in that PR except I removed the type signature because it required importing `Entity`.

Please let me know if there is a simpler way to fix this as I'm not adept at writing Python code.

### DIFF
--- a/custom_components/opensprinkler/__init__.py
+++ b/custom_components/opensprinkler/__init__.py
@@ -44,6 +44,14 @@ PLATFORMS = ["binary_sensor", "number", "select", "sensor", "switch", "text", "t
 TIMEOUT = 10
 
 
+def async_get_entities(hass: HomeAssistant):
+    """Get entities for a domain."""
+    entities = {}
+    for platform in async_get_platforms(hass, DOMAIN):
+        entities.update(platform.entities)
+    return entities
+
+
 async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
     """Set up OpenSprinkler from a config entry."""
     hass.data.setdefault(DOMAIN, {})
@@ -106,7 +114,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
     # Setup services
     async def _async_send_run_command(call: ServiceCall):
         await hass.helpers.service.entity_service_call(
-            async_get_platforms(hass, DOMAIN), SERVICE_RUN, call
+            async_get_entities(hass), SERVICE_RUN, call
         )
 
     hass.services.async_register(
@@ -118,7 +126,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
 
     async def _async_send_stop_command(call: ServiceCall):
         await hass.helpers.service.entity_service_call(
-            async_get_platforms(hass, DOMAIN), SERVICE_STOP, call
+            async_get_entities(hass), SERVICE_STOP, call
         )
 
     hass.services.async_register(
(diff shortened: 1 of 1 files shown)


curlydingo:
Just tested, looks good to me. I think this solution will do for now, though ultimately as @EdLeckert said previously we probably need a better approach to specifying these entities for services in the first place, since the current approach is a bit confusing on the GUI and is slow to execute.

This should close #278 too.

curlydingo (approved):

DylanGriffith:
I suppose we also need to consider version compatibility here.

This will only work with `2024.2.1` onwards. Meanwhile the previous code only works with `2024.1.2` and earlier.

Do we want to somehow build a version compatible with both?

Is there a way we can version this so that people won't be able to update this before updating home assistant? Maybe it doesn't matter which one they upgrade first because either way this is broken if they've updated HA before getting this update.

curlydingo:
> Is there a way we can version this so that people won't be able to update this before updating home assistant? Maybe it doesn't matter which one they upgrade first because either way this is broken if they've updated HA before getting this update.

We can update the content of hacs.json to require 2024.2.0? I've never actually used it before so might be a call for @vinteo though.

curlydingo:
Also would you mind fixing the linting? Couple of blank lines is all it needs.

Thanks

vinteo:
We can update in the hacs.json but it's not enforced. But I don't think it
is worth to maintain backwards compatibility, people just need to update

curlydingo:
Probably couldn’t hurt though? I suggest we update HACS.json and the version number and get it released soon, as I think it’s broken for everyone at the moment. 

I’m obviously pretty new here so I’ll leave releases to you unless you say otherwise.

vinteo:
yea sorry I mean let's do the hsac.json change.

EdLeckert (approved):
Seems from the discussion that this has been well thought out. While it's a small shame that it's not backward compatible, upgrading a component is a lot easier than rolling back HA.

Tested against `2024.2.1`

vinteo (approved):

vinteo:
Thank you @DylanGriffith

Issue #280, opened 2024-02-12
Inappropriate entities listed when configuring services


EdLeckert:
When using the `Developer tools\SERVICES` interface to learn about and test `OpenSprinkler` services, the user is presented with a confusing and usually inappropriate list of choices of entities that each service can utilize. The same applies in other locations in the UI, such as when using the `Automation Editor` to create a new automation.

The UI should present the user with only valid choices of entities when configuring an `OpenSprinkler` service.

vinteo:
fixed in #281

PR #281, opened 2024-02-12
Feat show valid service entities

EdLeckert:
Show only **valid** entities in the UI when configuring services. Fixes Issue #280 .

Services that affect the controller will list `switch.<controller>_enabled`.

Services that affect programs will list all `switch.<program>_program_enabled` entities.

Services that affect stations will list all `switch.<station>_station_enabled` entities.

Services that affect more than one of the above will list multiple types of entities.

The UI will not list values under `Choose area` or `Choose device`, as these would then select inappropriate entities for the service.

While other legitimate entities such as `sensor.<station>_station_status` could have also been listed, they provide no additional value and were left out to prevent clutter. However, nothing prevents a user from using these in `YAML`, so existing automations and scripts using these types of entities will continue to work.

services.yaml now uses the `target` keyword rather than `fields` to specify valid entities.

I've tested this against the examples in the docs, but additional testing by others would be appreciated.

### DIFF
--- a/custom_components/opensprinkler/services.yaml
+++ b/custom_components/opensprinkler/services.yaml
@@ -1,24 +1,31 @@
 run:
+  target:
+    entity:
+      device_class: 
+        - controller
+        - program
+        - station
+    device:
   fields:
-    entity_id:
-      example: "sensor.station_name"
     run_seconds:
       example: 60
     continue_running_stations:
       example: False
+
 stop:
-  fields:
-    entity_id:
-      example: "sensor.station_name"
+  target:
+    entity:
+      device_class: 
+        - controller
+        - station
+    device:
+
 set_water_level:
+  target:
+    entity:
+      device_class: controller
+    device:
   fields:
-    entity_id:
-      example: "sensor.opensprinkler_water_level"
-      required: true
-      selector:
-        entity:
-          integration: opensprinkler
-          domain: sensor
     water_level:
       example: 100
       required: true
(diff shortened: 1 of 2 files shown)


curlydingo (commented):
Looks good! Simplifies the experience a lot. I've tested all the services (against 2024.2.1) except for setting water level since it's the middle of summer here and I didn't want to forget to put it back. Really should figure out how to get a test environment set up for this sort of stuff.

I noticed that there's no selector for run_seconds in the run service. Technically a different issue but should we just add it here?

EdLeckert:
> Really should figure out how to get a test environment set up for this sort of stuff.

I use [VirtualBox](https://www.home-assistant.io/installation/windows/) on Windows. I only have the one physical OpenSprinkler controller, so I just disable the integration on my production system when testing with the virtual one.

> I noticed that there's no selector for run_seconds in the run service. Technically a different issue but should we just add it here?

Well, it _was_ there at one point in my testing. Now it's back. :smirk:

vinteo (approved):

curlydingo (approved):

curlydingo:
> 
> I use [VirtualBox](https://www.home-assistant.io/installation/windows/) on Windows. I only have the one physical OpenSprinkler controller, so I just disable the integration on my production system when testing with the virtual one.
> 
I find the HA devcontainer works really well for a test instance of the integration, but I'm hoping to find a way to run the OpenSprinkler firmware as a stub somewhere and avoid using my actual hardware. Relying on the hardware means I can't test all features, and I can only test when I don't have clothes out on the line :)

PR #286, opened 2024-05-06
Allow selfsigned certs

yeltnar:
This is to allow the form that creates the integration with the OpenSprinkler device to have a check box that will disable TLS (SSL) certificate checking so self signed certs can be used.

### DIFF
--- a/custom_components/opensprinkler/const.py
+++ b/custom_components/opensprinkler/const.py
@@ -12,7 +12,7 @@ CONF_PAUSE_SECONDS = "pause_duration"
 DOMAIN = "opensprinkler"
 
 DEFAULT_NAME = "OpenSprinkler"
-DEFAULT_CONF_VERIFY_SSL = True
+DEFAULT_VERIFY_SSL = True
 
 DEFAULT_SCAN_INTERVAL = 5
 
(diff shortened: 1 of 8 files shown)


custom_components/opensprinkler/strings.json
```diff
@@ -16,6 +16,7 @@
         "data": {
           "url": "URL",
           "password": "Password",
+          "verify_ssl": "Verify SSL",
```
vinteo, line 19: I feel like this is a bit vague, since it s kinda a related to security, it should be more explicit like:

- Allow self-signed SSL certificate
- Allow untrusted SSL certificate

For the other languages, if you feel Google translate doesn't cut it we can leave it as EN for now.

custom_components/opensprinkler/const.py
```diff
@@ -12,6 +12,7 @@
 DOMAIN = "opensprinkler"
 
 DEFAULT_NAME = "OpenSprinkler"
+DEFAULT_VERIFY_SSL = True
```
vinteo, line 15: Perhaps better to default to false

yeltnar: Are you thinking the default state should be allowing non trusted certificates? Just double checking.

vinteo:
Sorry I had it backwards. You are right it should be true. The checkbox
label should reflect that as well.

yeltnar:
I updated the English strings to be a little more telling. What do you think?

vinteo:
Looks good, if you could please fix the minor linting stuff, thanks

vinteo (approved):

vinteo:
Thank you very much

Issue #284, opened 2024-05-03
Run service in template switch doesn't work


cjhesse:
I'm not certain whether this is an issue with this integration, HASS, or my templating. That being said, the exact same service YAML works using the developer tools.

Here's what I have in `configuration.yaml`:

```
switch:
  - platform: template
    switches:
      sprinkler:
        friendly_name: Sprinkler
        value_template: >-
          {{ is_state('binary_sensor.z1_left_side_station_running', 'on') or
             is_state('binary_sensor.z2_back_main_station_running', 'on') or
             is_state('binary_sensor.z3_back_corner_station_running', 'on') or
             is_state('binary_sensor.z4_right_side_station_running', 'on') or
             is_state('binary_sensor.z5_front_garden_station_running', 'on') or
             is_state('binary_sensor.z6_front_main_station_running', 'on') }}
        availability_template: >-
          {{ has_value('binary_sensor.z1_left_side_station_running', 'on') or
             has_value('binary_sensor.z2_back_main_station_running', 'on') or
             has_value('binary_sensor.z3_back_corner_station_running', 'on') or
             has_value('binary_sensor.z4_right_side_station_running', 'on') or
             has_value('binary_sensor.z5_front_garden_station_running', 'on') or
             has_value('binary_sensor.z6_front_main_station_running', 'on') }}
        turn_on:
          service: opensprinkler.run
          data:
            entity_id: switch.opensprinkler_enabled
            run_seconds:
              0: 60
              1: 60
              2: 60
              3: 60
              4: 60
              5: 60
            continue_running_stations: true
        turn_off:
          service: opensprinkler.stop
          data:
            entity_id: switch.opensprinkler_enabled
```

When turning the switch on, nothing (apparently) happens. I didn't see any warnings or errors in the log that appeared related.

Turning the switch off does successfully stop the sprinkler when it's already running.

I tested the same "turn on" action in Developer tools > Services, and that worked:

```
service: opensprinkler.run
data:
  entity_id: switch.opensprinkler_enabled
  run_seconds:
    0: 60
    1: 60
    2: 60
    3: 60
    4: 60
    5: 60
  continue_running_stations: true
```

EdLeckert:
The only problem I see with your template is that `has_value` only takes one parameter, but that's not relevant to the failure.

[...]

PR #288, opened 2024-05-30
fix: allow run_seconds to use int or str as keys

EdLeckert:
This resolves Issue #284, which pointed out that one of the formats for creating template switches used in the examples no longer works.

Consider this snippet:

```
  run_seconds:
    0: 60
    1: 60
    2: 60
    3: 60
    4: 60
    5: 60
```

If this is called as part of `opensprinkler.run` from `Developer tools > Services`, the keys are converted to strings, which is what the code expects. However, if it is part of a template switch, they arrive as integers, which `dict.get()` interprets as a lookup failure and returns `None`. Ultimately, the run durations are set to 0.

This PR attempts to find both int and str keys. If you know of a better way to cover both cases than to use the `default` parameter of `dict.get()`, please speak up.

### DIFF
--- a/custom_components/opensprinkler/__init__.py
+++ b/custom_components/opensprinkler/__init__.py
@@ -316,9 +316,12 @@ class OpenSprinklerControllerEntity:
         if isinstance(run_seconds, dict):
             run_seconds_list = []
             for _, station in self._controller.stations.items():
+                seconds = run_seconds.get(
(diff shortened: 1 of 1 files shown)


vinteo (approved):

EdLeckert:
@vinteo: I see you have approved this PR. Do you want me to merge it?

Issue #293, opened 2024-06-30
Incorrect unavailable handling


bob-tm:
Hi,

I have regular blackouts and integration do not correctly handle it. 

![logbook](https://github.com/vinteo/hass-opensprinkler/assets/47008924/33b14a32-25f4-419a-8c70-b5f50edce8ac)
Let's look to an example
8.12.30 - blackout 
8.12.44 - it's unavailable  with "Timeout fetching OpenSprinkler resource status data" message. 
8.13.28 - it's alive with "Fetching OpenSprinkler resource status data recovered", but it's impossible. Still zero power. 
![log1](https://github.com/vinteo/hass-opensprinkler/assets/47008924/651e7854-4523-4af3-b5db-c802eb286531)

"Timeout fetching OpenSprinkler resource status data" it's very rare message. When I saw it - unavailable handled correctly.  But when there are: 
"Finished fetching OpenSprinkler resource status data in 10.003 seconds (success: False)",
"Cannot connect to controller",
everything is showed as "online"

Second strange thing: there is 10 sec timeouts with 5 seconds SCAN_INTERVAL.  Looks like it should be less than SCAN_INTERVAL. 
![log2](https://github.com/vinteo/hass-opensprinkler/assets/47008924/1cd232a8-36de-4152-996d-52ab5c084e12)
Another situation with "Fetching OpenSprinkler resource status data recovered" while there is total blackout

My current dirty solution - full reload OpenSprinkler integration on each blackout.

![log3](https://github.com/vinteo/hass-opensprinkler/assets/47008924/fe9f3d13-73ac-40aa-b0f7-f61a800fdc2d)
Now it's powered. Example in logs for short unavailable case. 
For this case I suggest count timeouts. For example if there is two continues timeouts - then change availability attribute or reset timeout counter.

bob-tm:
Seems like reducing timeout from 10 sec to 3sec helps with unavailable  while blackout.

functionpointer:
My OpenSprinkler got OOM killed. The integration remained mostly available, despite numerous errors:
```
Aug 01 09:40:00 netberry hass[418]: 2024-08-01 09:40:00.779 ERROR (MainThread) [custom_components.opensprinkler] Cannot connect to controller
[...]

PR #298, opened 2024-08-01
Fix going unavailable

functionpointer:
Fixes https://github.com/vinteo/hass-opensprinkler/issues/293

After investigating, i think the problem isn't really the `TIMEOUT` nor `SCAN_INTERVAL`, as this is something HAs built-in `DataUpdateCoordinator` should handle.

Instead, the problem is error management. We do get an error, but we ignore it!
Instead, we rely on `controller._state`, but this doesn't get cleared when an error occurs.

This PR fixes the problem by propagating the error if we get one.


functionpointer:
The `black` formatter seems to be outdated, as my local version (24.4.2), formats slightly differently.

vinteo:
Thanks for this

I updated the linter can you please rebase with master

vinteo (approved):

vinteo:
Thank you

PR #301, opened 2024-08-02
Add support for reauth flow

functionpointer:
With https://github.com/vinteo/hass-opensprinkler/pull/298 merged, we should also add support for the reauthentication flow.

This PR also simplifies exception management on initial poll, by using HA's built-in `async_config_entry_first_refresh()` function.
It will automatically raise `ConfigEntryNotReady` when an error occurs. This allows us to remove a big `try:` block.

### DIFF
--- a/custom_components/opensprinkler/__init__.py
+++ b/custom_components/opensprinkler/__init__.py
@@ -15,7 +15,6 @@ from homeassistant.const import (
     CONF_VERIFY_SSL,
 )
 from homeassistant.core import HomeAssistant, ServiceCall
-from homeassistant.exceptions import ConfigEntryNotReady
 from homeassistant.helpers import config_validation as cv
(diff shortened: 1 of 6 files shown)


EdLeckert (approved):
Seems to work fine. The result reason displayed in the dialog by `async_abort` seems a little rough, but I see that it is fairly common practice to do it this way.

image

Another thing I noticed that has probably been this way for a while is that a non-existent URL does not trigger an `InvalidURL` exception. Instead, it results in a timeout that displays as `Unexpected error`. Might want to look into that at some point.

vinteo:
the rough message seems to be due to missing strings in the localisation? example:

https://github.com/home-assistant/core/blob/4fd92c17f077bf842d054a48c2af6ecc48722b74/homeassistant/components/androidtv_remote/strings.json#L35

vinteo (approved):

vinteo:
thanks again

Issue #313, opened 2024-10-01
OpenSprinklerStationEntity.run() got an unexpected keyword argument 'continue_running_stations'


paranerd:
I set up a Template Switch using the UI. It has an option to let stations run in parallel when being started manually. However the `run()` function doesn't actually seem to expect the option.

I'm assuming either the frontend or backend is wrong here. Does anyone have a clue?

EdLeckert:
First of all, do not confuse `continue_running_stations` with running stations in parallel. All this parameter does is not stop already running stations--kind of. Depending on the order of the already running stations vs. the new stations, an already running station may be paused and pre-empted by the new one, but will be queued to continue later with whatever time is remaining. This is because the _new_ list of run-times for _all_ stations must be submitted in station order, so they will then run in station order.

In any case, what you have encountered is that the `run` service (action) takes different parameters depending on the chosen target entity. The `continue_running_stations` parameter only works when a controller is the target--not a program or station--and then only if index/seconds pairs are used.

I have submitted PR #314 to try to alleviate the confusion by eliminating the display of parameters that are not valid. It's a work in progress. In the meantime, leave off that parameter unless targeting a controller.

paranerd:
Thanks for the in depth analysis!
I can confirm that my stations can run in parallel regardless of the setting.

PR #314, opened 2024-10-05
Split run services by target type

EdLeckert:
This is a continuation of an earlier effort to simplify and clarify action (service) calling. That effort limited the available entities presented by the UI to only those appropriate to the action. This PR was inspired by Issue #313 and is intended to reduce the available selection of parameters available in the UI to the `run` action depending on the use case.

It has become clear that the only way to effectively do this was to split the `run` action into multiple actions, one for each type of target entity. This is due to the fact that each entity type (controller, station, and program) takes a different set of parameters: `run_seconds` and `continue_running_stations`, only `run_seconds`, and no parameters, respectively.

Therefore, I have added these new actions:

- `run_once`
- `run_station`
- `run_program`

I have left the `run` service as-is (with one exception--see below), with the hope that it can be deprecated and removed in the future.

I also noticed that the logic for `continue_running_stations` was reversed, so that has been corrected. This could be considered a breaking change, but is the only one.

Finally, I have updated the docs (except for non-"en" translations), probably more than I had intended, but I wanted the existing examples to be consistent with the new ones.

I realize this is a bit to take in. Let me know your thoughts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,9 @@ Last tested on OS API `2.2.0` and Home Assistant `2024.2.0`
 - Sensors for water level, last runtime and rain delay stop time
 - Switches for each program and station to enable/disable program or station
 - Switch to enable/disable OpenSprinkler controller operation
-- Services to run and stop stations
-- Service to run programs
-- Services to pause, set a rain delay, and set the water level.
+- Actions to run and stop stations
+- Action to run programs
+- Actions to pause, set a rain delay, and set the water level.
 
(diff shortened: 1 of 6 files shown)


vinteo (approved):

vinteo:
@EdLeckert could you do me a favour and write a small brief about the breaking change that I can put into the release notes. Thank you

EdLeckert:
**Breaking Change**: The logic of the `run` command parameter `continue_running_stations` has been corrected. Previously, if the value of the parameter was set to `true`, it would **_not_** continue running stations that were already running, and if set to `false`, it would.

PR #359, opened 2026-05-01
feat: add use_weather_adjustment, queue_option, and shift_sequential_stations parameters

jan-auer:
Adds three new optional parameters to run and stop actions, backed by firmware API parameters `uwt`, `qo`, and `ssta`. Requires vinteo/py-opensprinkler#112.

**`use_weather_adjustment`** (`run_once`, `run_program`)
Boolean. When `true`, the controller's current weather-based water level adjustment is applied to run durations. When `false`, stations run at 100% regardless of weather adjustment. Maps to firmware parameter `uwt`.

**`queue_option`** (`run_once`, `run_program`, `run_station`)
String. Controls how the run is placed in the station queue. Maps to firmware parameter `qo`.
- `append` — add to the end of the queue
- `preempt` — insert at the front of the queue
- `replace` — clear the queue and start fresh (`run_once` and `run_program` only)

**`shift_sequential_stations`** (`stop`, station entities only)
Boolean. When `true`, remaining stations in the same sequential group shift forward when a station is stopped early. Maps to firmware parameter `ssta`.

All parameters are optional and default to `None`, which omits them from the API call and lets firmware defaults apply.


custom_components/opensprinkler/__init__.py
```diff
@@ -386,15 +410,15 @@ async def run(self, run_seconds=None, continue_running_stations=None):
...
             await self._coordinator.async_request_refresh()
             return
 
-        await self._controller.run_once_program([int(s) for s in run_seconds])
+        await self._controller.run_once_program([int(s) for s in run_seconds], **kwargs)
         await self._coordinator.async_request_refresh()
         return
 
-    async def stop(self):
+    async def stop(self, shift_sequential_stations=None):
```
vinteo, line 418: this is not used

jan-auer: The underlying API to stop all stations doesn't have the `ssta` parameter, but IIUC we need to add the parameter to the signature for stations since there's just a single stop service/action exposed. We could introduce a new `opensprinkler.stop_station` action that has this new parameter, but I thought it might be nicer to stick to a single action.

To confirm, let me know if you prefer a separate action and I'll update.

vinteo:
Thank you, please take a look at some of the linting stuff, also point to the pyopensprinkler 0.7.16 version

vinteo (approved):

EdLeckert (changes requested):
Two problems I see:

1) `run_once_program` in `py-opensprinkler/__init__.py` does not support the new parameters, so calling `opensprinkler.run_once` with them results in:

`TypeError: Controller.run_once_program() got an unexpected keyword argument 'uwt'`

I have no idea how the tests passed. I'm looking at `py-opensprinkler` `master` and the only place I see `uwt` is in `program.py`. Perhaps I'm missing something, but...

2) The new action parameter definitions are not included in any language translations, so they show up in `Developer tools/Actions` as `use_weather_adjustment` instead of `Use weather adjustment` with its description, for example. `strings.json` should be copied over `en.json` so at least English is covered. Perhaps the translators of the other language files can be pinged when this is complete?

jan-auer:
Thank you, my bad. The patch to py-opensprinkler was incomplete and the tests are mocking all of the functions. This first needs a fix in py-opensprinkler, then I'll follow up with the rest here and translations.

jan-auer:
Created https://github.com/vinteo/py-opensprinkler/pull/114 and updated translations. 

There were a few translations missing in Slovak and German that I added now and I took the liberty to fix up some wrong German translations along the way. However, I don't speak Portuguese or Slovak, so these translations were done online.  

Unfortunately, I do not have a setup to test this e2e against an opensprinkler simulator/demo instance. May I ask one of you to help out with this?

EdLeckert:
Where are the translation changes to `hass-opensprinkler`?

EdLeckert (approved):
Approved pending release of py-opensprinkler changes. Not sure if a version bump of that is required, which would require a manifest change here.

vinteo:
released pyopensprinkler 0.7.17, please update again

https://pypi.org/project/pyopensprinkler/0.7.17/

vinteo:
Thank you, released in v1.5.5

jan-auer:
Thanks everyone!

PR #371, opened 2026-07-25
Feature: Support recently added scheduling features

EdLeckert:
Additional scheduling options have been added to the firmware in the last several releases. The `Single-run` and `Monthly` options share the same locations in the message structure with the currently supported `Weekly` and `Interval-day` options, and if these new options are chosen in the `OpenSprinkler` UI, the integration will not interpret the data correctly. This PR corrects this issue by adding support for all four scheduling options.

Also added is the ability to limit scheduling to a date range.

Part of the inspiration for this work is the intention of soon adding a reliable `Next Run` feature for each program and for the system as a whole. It is necessary for the integration to be aware of all possible scheduling parameters for this feature to be reliable.

### DIFF
--- a/custom_components/opensprinkler/__init__.py
+++ b/custom_components/opensprinkler/__init__.py
@@ -59,7 +59,16 @@ from .const import (
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


vinteo (approved):

vinteo:
released in v1.6.0

PR #377, opened 2026-08-12
Expose station group and controller station delay as attributes

clarejor:
Station sequencing is not observable from Home Assistant today. Which stations serialize with each other, and which run concurrently, is decided by the station group id, and the gap the controller inserts between sequential stations is decided by the station delay option. Neither value reaches an entity, so an automation or an external scheduler cannot work out when a chain of stations will finish.

Changes:

- `manifest.json`: bump the requirement to `pyopensprinkler==0.7.20`, the release that added `Station.group` and `Controller.station_delay`. The `version` field is left alone.
- `OpenSprinklerStationEntity`: add `group` to the station attributes. The raw `stn_grp` id is passed through, so a consumer can compare two stations for membership of the same sequential group. 255 means the parallel group and is passed through as 255. A derived boolean was considered and rejected: stations in two different sequential groups also run concurrently, which a boolean cannot express.
- `ControllerOperationSwitch`: add `station_delay` to the controller attributes. Seconds, negative meaning stations overlap.

`Station.group` raises `FirmwareNotSupportedError` below firmware v2.2.0(1). The attribute is added to the existing per attribute try/except loop, so on older firmware the key is simply absent, exactly as other unsupported attributes already behave, and the rest of the entity's attributes are unaffected. `Controller.station_delay` needs no firmware gate and returns None when the option is missing.

Tests in `tests/test_station_attributes.py` cover the supported case, the parallel sentinel, omission of the key when the property raises, that a group change is reflected on the next state write rather than cached, and both signs of the station delay.

### DIFF
--- a/custom_components/opensprinkler/__init__.py
+++ b/custom_components/opensprinkler/__init__.py
@@ -541,6 +541,7 @@ class OpenSprinklerStationEntity:
             "index",
             "is_master",
             "running_program_id",
+            "group",
         ]:
             try:
                 attributes[attr] = getattr(self._station, attr)
--- a/custom_components/opensprinkler/manifest.json
+++ b/custom_components/opensprinkler/manifest.json
@@ -7,6 +7,6 @@
   "documentation": "https://github.com/vinteo/hass-opensprinkler",
   "iot_class": "local_polling",
   "issue_tracker": "https://github.com/vinteo/hass-opensprinkler/issues",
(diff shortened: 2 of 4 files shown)


vinteo (approved):

EdLeckert (commented):
Looks good. This PR should go before mine, since I've bumped the version to 2.0.0 and added a `requirements` item.

Issue #163, opened 2022-02-25
change url


mac-city:
is it possible to change the url on my up and running setup?  I changed wifi and ip address and just want to change it without having to delete the integration and probably losing the setup.  love this integration!  thanks!

vinteo:
no, I don't think it is possible

mac-city:
bummer

travisghansen:
You can hack the file in the storage dir pretty easy yeah. Just shut down ha, edit the file, and restart.

mac-city:
oh shoot, I just removed, reinstalled.  this has come up on other integrations so let me see if I can find it for future reference, thanks!  so I would have to ssh to get there?  where is the storage directory.  I don't think I get get to it if ha is off?  I am running home assistant on proxmox.  the only way I know how to shut down ha is to turn the machine off, and then I couldn't get to that?

travisghansen:
I'm not sure about how to get there in the os. The file you're looking for is `<config dir>/.storage/core.config_entries`

PR #386, opened 2026-08-27
Add Reconfigure flow to change controller URL

### LINKED ISSUE #385: Add Reconfigure flow to change controller URL
## Problem

When an OpenSprinkler controller’s IP address or hostname changes, there is no way to update the stored URL without removing and re-adding the integration. That can create a new config entry and new device IDs, which breaks existing automations and history.

This was requested in #163 (2022). At the time, Home Assistant did not yet have a Reconfigure flow, and the maintainer correctly said it was not possible.

## Proposal

I intend to submit a PR that adds a config-entry **Reconfigure** flow using the current Home Assistant helpers (minimum HA **2024.11.0**):

- Settings → Devices & services → OpenSprinkler → ⋮ → Reconfigure
- First form: **Controller URL** and **Verify SSL** only
- Test the new URL with the **existing stored password**
- If the controller API returns a MAC: validate it matches the existing config entry unique ID, then update that same entry in place
- If the API does not return a MAC (legacy firmware): second step `reconfigure_mac`; the entered MAC must match the existing unique ID
- On success: `async_update_reload_and_abort(..., data_updates=...)` so the same config entry, unique ID, device IDs, and entity IDs are preserved
- Does **not** create a second config entry
- Does **not** change unique-ID construction (MAC-based)

This matches Home Assistant’s current guidance that Reconfigure is the mechanism for changing required setup data (such as host/URL), and that a successful reconfigure should update the existing entry rather than create another.

## Out of scope for this PR

- Duplicate-add / DHCP auto-update on the initial setup flow
- A password field on Reconfigure
- Options flow / scan interval UI
- Simultaneous controller URL + password changes (unsupported in this PR; see Known Limitations)

## Known Limitations

Simultaneous controller URL + password changes are unsupported in this PR.

## References

- #163
- https://developers.home-assistant.io/docs/core/integration/config_flow/#reconfigure
- https://developers.home-assistant.io/blog/2024/10/21/reauth-reconfigure-helpers/

I am willing to submit the PR.

toddhutch:
## Summary
- Adds a Home Assistant **Reconfigure** flow so the OpenSprinkler controller URL (IP or hostname) and Verify SSL setting can be changed on the existing config entry. Device IDs, entity IDs, and automations stay the same.
- First form is URL + Verify SSL only. The stored password is reused. If the API returns a MAC, it must match the existing unique ID (`wrong_device` otherwise). If the API has no MAC (legacy firmware), a second `reconfigure_mac` step asks for the MAC and requires the same match.
- Raises the HACS minimum to Home Assistant **2024.11.0** and adds a pinned CI job for that version, because the flow uses the 2024.11 reconfigure helpers.
- Closes #385. This is the current-HA way to do what #163 asked for in 2022.

**Out of scope:** duplicate-add / DHCP auto-update, a password field on Reconfigure, Options flow, and Reauth modernization.

**Known limitation:** Simultaneous controller URL + password changes are unsupported in this PR; see Known Limitations.

## Test plan
- [x] Local: `homeassistant==2024.11.0` + `pytest-homeassistant-custom-component==0.13.181` on Python 3.12 — 48 tests passed (9 new config-flow cases + existing suite)
- [ ] CI: existing unpinned Tests job still green
- [ ] CI: new `Home Assistant 2024.11 compatibility` job green
- [ ] Live (later, with the reporter; do **not** uninstall the integration): copy `custom_components/opensprinkler` over the installed integration, then Reconfigure when the old URL is already dead, and confirm device/entity IDs are unchanged
- [ ] Confirm pointing Reconfigure at a *different* controller is rejected
- [ ] Confirm a URL change with a *new* password is not claimed to work in this PR

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,3 +36,28 @@ jobs:
       - name: Run tests
         run: |
           pytest tests/ -v --tb=short
+
+  test-ha-2024-11:
+    name: Home Assistant 2024.11 compatibility
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v7
+
+      - name: Set up Python 3.12
+        uses: actions/setup-python@v7
+        with:
+          python-version: "3.12"
+
+      - name: Install pinned test dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements-test-ha-2024.11.txt
+
+      - name: Run tests
+        run: |
+          pytest tests/ -v --tb=short
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -54,9 +54,9 @@ jobs:
           python-version: "3.12"
 
       - name: Install pinned test dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install -r requirements-test-ha-2024.11.txt
+        # pyopensprinkler and mock-open (via pytest-homeassistant-custom-component)
+        # ship sdists only, so --only-binary :all: cannot be used here.
+        run: python -m pip install homeassistant==2024.11.0 pytest-homeassistant-custom-component==0.13.181 pyopensprinkler==0.7.20 josepy==1.14.0 # NOSONAR
 
       - name: Run tests
         run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,7 +38,7 @@ jobs:
           pytest tests/ -v --tb=short
 
   test-ha-2024-11:
-    name: Home Assistant 2024.11 compatibility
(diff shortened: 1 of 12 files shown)


toddhutch:
Live-tested on a production Home Assistant install.

Copied `custom_components/opensprinkler` over the existing integration (did not uninstall). Reconfigured after the old controller URL was already unreachable. The config entry and entity IDs stayed the same, and the controller came back on the new URL.

EdLeckert (changes requested):
Feature works as described. However, found two minor issues:

- SonarCloud alert about 'useless "yield"' in tests/conftest.py
- Reference to `Configuration` then `Integrations` in README.md, rather than `Settings`...

toddhutch:
Addressed both review items:

- Removed the unused `yield` from `tests/conftest.py`.
- Updated the README to `Settings` then `Devices & services`.

PR #394, opened 2026-09-04
Relax suntime dependency

### LINKED ISSUE #393: Support for suntime == 1.4.0?
I see we're pinned to suntime 1.3.2. Is there any reason for the hyper specific pin? AFAICT, there are no breaking changes in suntime 1.4.0: https://github.com/SatAgro/suntime/releases/tag/v1.4.0

Asking because I maintain the [`home-assistant-custom-components.opensprinkler` package](https://search.nixos.org/packages?channel=unstable&query=home-assistant-custom-components.opensprinkler) in nixpkgs. We package the latest version of suntime, which doesn't play nicely with the super strict bound here. I can workaround the issue, but thought you might appreciate a heads up. 

`<Possibly unwanted opinions follow>`
In my experience in the python ecosystem, "pessimistic" bounds like this cause more problems than they're worth. As python only allows one version of a give package, if another package wants a different version of suntime, we're hosed. For a long writeup about this, I recommend <https://iscinumpy.dev/post/bound-version-constraints/#tldr>. tl;dr:

> "Libraries/packages should be setting a floor, and if necessary excluding known buggy versions, but otherwise don’t cap the maximum version as you can’t predict future compatibility"
> -Python Steering Council Member and packaging expert Brett Cannon

jfly:
There are no known issues with newer versions of suntime, and it's generally frowned upon to be pessimistic about your dependencies in the Python ecosystem.

This fixes https://github.com/vinteo/hass-opensprinkler/issues/393

### DIFF
--- a/custom_components/opensprinkler/manifest.json
+++ b/custom_components/opensprinkler/manifest.json
@@ -7,6 +7,6 @@
   "documentation": "https://github.com/vinteo/hass-opensprinkler",
   "iot_class": "local_polling",
   "issue_tracker": "https://github.com/vinteo/hass-opensprinkler/issues",
-  "requirements": ["pyopensprinkler==0.7.22", "suntime==1.3.2"],
+  "requirements": ["pyopensprinkler==0.7.22", "suntime>=1.3.2"],
   "version": "2.2.0"
 }


vinteo:
thank you

